PR #5, opened 2019-09-18
Fix crash in GC: PyTuple_SET_ITEM steals reference

iakov:
Py_Finalize crash is caused by references mishandling in Enums. This PR fixes the issue.


sboehmann:
@iakov Thank you for the patch. Looks good and is merged now.

f03el:
Sorry to necropost, but what happened to commit 642b73e? I pulled from this repo today, and while I had that merge commit locally from earlier pulls, it seems to have vanished from this repo.

iakov:
At least the fix is still in the upstream's master. :) https://github.com/MeVisLab/pythonqt/commit/963b667acc40b30fe6d223725730db697739e115

f03el:
Yeah, but if a commit exists in a branch, it's not expected to disappear since that messes up everybody's Git history.

PR #16, opened 2020-08-09
Fix memory leak

iakov:
The last element was not deleted even when it is the single one in the list.


florianlink:
Thanks for the fix!

PR #24, opened 2020-10-09
Fix reference/memory leaks

gregor-anich-uibk:
Tested with Python-3.8.5 debug build and memcheck, does not fix all leaks but some, reduces dynamic memory leaks by 2/3rds

### DIFF
--- a/build/python.prf
+++ b/build/python.prf
@@ -2,10 +2,19 @@
 
 
 # Change this variable to your python version (2.6, 2.7, 3.3, ...)
-PYTHON_VERSION=$$(PYTHON_VERSION)
+isEmpty( PYTHON_VERSION ) {
+  PYTHON_VERSION=$$(PYTHON_VERSION)
+}
 isEmpty( PYTHON_VERSION ) {
   win32:PYTHON_VERSION=27
   unix:PYTHON_VERSION=2.7
+}
+
+isEmpty( PYTHON_DIR ) {
+  PYTHON_DIR=$$(PYTHON_DIR)
+}
+!isEmpty( PYTHON_DIR ) {
+  PYTHON_DIR=$${PYTHON_DIR}/
 }
 
 message(Using Python version $${PYTHON_VERSION})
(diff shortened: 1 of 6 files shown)


florianlink (commented):
Looks pretty good and you seem to know your business! I always hated these Api differences (which call steals and which doesn't). I will ask another colleague to have a look as well and then we can merge this.

src/PythonQt.cpp
```diff
@@ -525,6 +536,7 @@ void PythonQtPrivate::createPythonQtClassWrapper(PythonQtClassInfo* info, const
     PythonQtClassInfo* outerClassInfo = lookupClassInfoAndCreateIfNotPresent(outerClass);
     outerClassInfo->addNestedClass(info);
   } else {
+    Py_INCREF(pyobj);
     PyModule_AddObject(pack, info->className(), pyobj);
```
usiems, line 540: PyModule_AddObject only decrements the reference count on success, so you should check the return code.

src/PythonQt.cpp
```diff
@@ -1048,24 +1068,30 @@ PythonQtObjectPtr PythonQt::createUniqueModule()
 
 void PythonQt::addObject(PyObject* object, const QString& name, QObject* qObject)
 {
+  PyObject *wrappedObject = _p->wrapQObject(qObject);
   if (PyModule_Check(object)) {
-    PyModule_AddObject(object, QStringToPythonCharPointer(name), _p->wrapQObject(qObject));
+    Py_XINCREF(wrappedObject);
+    PyModule_AddObject(object, QStringToPythonCharPointer(name), wrappedObject);
```
usiems, line 1074: PyModule_AddObject only decrements the reference count on success, so you should check the return code.

src/PythonQt.cpp
```diff
@@ -2105,10 +2133,12 @@ const QMetaObject* PythonQtPrivate::buildDynamicMetaObject(PythonQtClassWrapper*
   builder.setClassName(((PyTypeObject*)type)->tp_name);
 
   PyObject* dict = ((PyTypeObject*)type)->tp_dict;
-  Py_ssize_t pos = NULL;
+  Py_ssize_t pos = 0;
   PyObject* value = NULL;
   PyObject* key = NULL;
-  static PyObject* qtSlots = PyString_FromString("_qtSlots");
+  static PyObject* qtSlots = NULL;
+  if (!qtSlots)
```
usiems, line 2140: Please use braces even for one-line blocks

src/PythonQtImporter.cpp
```diff
@@ -289,14 +289,17 @@ PythonQtImporter_load_module(PyObject *obj, PyObject *args)
...
+      err = PyDict_SetItemString(dict, "__package__", fullnameObj);
+      Py_XDECREF(fullnameObj);
       if (err != 0) {
         Py_DECREF(code);
         Py_DECREF(mod);
         return NULL;
       }
 #endif
     }
+    Py_DECREF(mod);
```
usiems, line 302: Is this correct? The documentation states that PyModule_AddModule returns a borrowed reference. But I see that Py_DECREF is also used on mod before, so it seems as if this is the correct way to do this...

usiems (commented):
Please have a look at the inline comments.

src/PythonQt.cpp
gregor-anich-uibk: Yes, I saw this, there are many more such calls and I thought i leave it up for the next round of fixing, but now I fixed all of them by adding a wrapper function to do the job.

src/PythonQt.cpp
gregor-anich-uibk: Done.

src/PythonQtImporter.cpp
gregor-anich-uibk: Yes, good catch, I probably wrongly assumed the existing code was correct but you are right, documentation says it returns a borrowed reference so I removed all the DECREFs.

build/python.prf
```diff
@@ -8,41 +8,43 @@ isEmpty( PYTHON_VERSION ) {
...
+  # on linux, python-config is used to autodetect Python.
+  # make sure that you have installed a matching python-dev package.
   
   system(python$${PYTHON_VERSION}-config --embed --libs) {
     unix:LIBS += $$system(python$${PYTHON_VERSION}-config --embed --libs)
   } else: unix:LIBS += $$system(python$${PYTHON_VERSION}-config --libs)  
-  unix:QMAKE_CXXFLAGS += $$system(python$${PYTHON_VERSION}-config --includes)
-}
+  unix:QMAKE_CXXFLAGS += $$system(python$${PYTHON_VERSION}-config --includes)
+  unix:QMAKE_LFLAGS += $$system(python$${PYTHON_VERSION}-config --ldflags)
```
gregor-anich-uibk, line 49: I had to add this line to build against Python-3.8.5 on my OpenSuSE machine

gregor-anich-uibk:
I tried to address the issues you showed up, please let me know if it's now fine to be committed. I wish I had the time to do a proper fixup/cleanup of PythonQt, but I found that Python itself cannot really clean up all the memory which surprised me, it's a nice language and I really like to use it for scripting my application, but I'll have to live with small leaks :smile:

usiems:
I have no objections to this merge request. Florian?

florianlink:
Yes, looks good!

gregor-anich-uibk:
Nice, thanks, will contribute again if I do more fixes!

PR #25, opened 2020-10-19
Fix build, fix potential crash

gregor-anich-uibk:
Sorry for breaking the old build system :disappointed: 

The changes to PythonQt(Safe)ObjectPtr are only needed if one calls Py_Finalize() I believe. In this case the destructor calls Py_DECREF after the python interpreter was destroyed. For "normal" users this should never happen, but still shouldn't hurt to have this check and safe others time when debugging their code.


usiems (commented):
I'm a little bit curious what the cause of the crash was in PythonQtObjectPtr::~PythonQtObjectPtr. The PythonQtObjectPtr destructors didn't use PythonQt::self(), as far as I can see, only Python itself. What should fail are the access methods like call() etc., which didn't get any validity checks.

src/PythonQtConversion.cpp
```diff
@@ -998,8 +999,11 @@ QVariant PythonQtConv::PyObjToQVariant(PyObject* val, int type)
     } else if (PyLong_Check(val)) {
       // return int if the value fits into that range,
       // otherwise it would not be possible to get an int from Python 3
-      qint64 d = PyLong_AsLongLong(val);
-      if (d > std::numeric_limits<int>::max() ||
+      int overflow = 0;
+      qint64 d = PyLong_AsLongLongAndOverflow(val, &overflow);
+      if (overflow != 0) {
+        type = QVariant::ByteArray; // if it's too long for long long return byte array
```
usiems, line 1005: Setting the type to QVariant::ByteArray leads to the use of PythonQtConv::PyObjGetBytes later on, which only works if PyBytes_Check returns true. So I guess this doesn't do anything useful.

src/PythonQtConversion.cpp
```diff
@@ -947,7 +947,8 @@ double PythonQtConv::PyObjGetDouble(PyObject* val, bool strict, bool &ok) {
 template <typename Map> 
 void PythonQtConv::pythonToMapVariant(PyObject* val, QVariant& result)
 {
-  if (PyMapping_Check(val)) {
+  // check if val is not a string, otherwise AttributeError("'str' object has no attribute 'items'") is raised which is misleading
```
usiems, line 950: I believe this can not just occur for strings, but for other types too. Perhaps it would be better to just clear the error produced by PyMapping_Items instead. (If an error is already set we could skip the whole thing to avoid clearing an external error.)

usiems (changes requested):

mrbean-bremen:
Closing as stale.

PR #27, opened 2020-10-30
Update python.prf

Jerrfey:
Fix a few typos

### DIFF
--- a/build/python.prf
+++ b/build/python.prf
@@ -3,7 +3,7 @@
 
 # Change this variable to your python version (2.6, 2.7, 3.3, ...)
 isEmpty( PYTHON_VERSION ) {
-  PYTHON_VERSION=$$(PYTHON_VERSION)
+  PYTHON_VERSION=$${PYTHON_VERSION}
 }
 isEmpty( PYTHON_VERSION ) {
   win32:PYTHON_VERSION=27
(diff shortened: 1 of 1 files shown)


build/python.prf
```diff
@@ -3,15 +3,15 @@
 
 # Change this variable to your python version (2.6, 2.7, 3.3, ...)
 isEmpty( PYTHON_VERSION ) {
-  PYTHON_VERSION=$$(PYTHON_VERSION)
+  PYTHON_VERSION=$${PYTHON_VERSION}
```
usiems, line 6: I believe this is wrong. According to the qmake documentation (https://doc.qt.io/qt-5/qmake-language.html) $$() is used to access environment variables, and $${} to access qmake variable ($${PYTHON_VERSION} is almost the same as $$PYTHON_VERSION, but allowing concatenation). And assigning the variable to itself doesn't make sense.

build/python.prf
```diff
@@ -3,15 +3,15 @@
...
 }
 isEmpty( PYTHON_VERSION ) {
   win32:PYTHON_VERSION=27
   unix:PYTHON_VERSION=2.7
 }
 
 isEmpty( PYTHON_DIR ) {
-  PYTHON_DIR=$$(PYTHON_DIR)
+  PYTHON_DIR=$${PYTHON_DIR}
```
usiems, line 14: Same as above.

usiems (commented):
Sorry, I believe these changes should not be applied.

PR #43, opened 2021-04-13
Update __spec__ too (if existing) when setting the import path

usiems:
This change avoids a warning message "__package__ != __spec__.parent" (in Python 3.9) on relative imports from modules which have the import paths set with PythonQt::self()->setModuleImportPath(). Since __spec__ is meant to replicate some settings like __path__, this seems the correct solution.

### DIFF
--- a/src/PythonQt.cpp
+++ b/src/PythonQt.cpp
@@ -1680,6 +1680,20 @@ void PythonQt::overwriteSysPath(const QStringList& paths)
 void PythonQt::setModuleImportPath(PyObject* module, const QStringList& paths)
 {
   PyModule_AddObject(module, "__path__", PythonQtConv::QStringListToPyList(paths));
+#if PY_MAJOR_VERSION > 3 || (PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION >= 4)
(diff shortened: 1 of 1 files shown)


benbarate (approved):

src/PythonQt.cpp
```diff
@@ -1713,6 +1713,20 @@ void PythonQt::overwriteSysPath(const QStringList& paths)
 void PythonQt::setModuleImportPath(PyObject* module, const QStringList& paths)
 {
   PyModule_AddObject_DECREF(module, "__path__", PythonQtConv::QStringListToPyList(paths));
+#if PY_MAJOR_VERSION >= 3 || (PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION >= 4)
+  // Since Python 3.4 a module has a __spec__ member of type ModuleSpec which
```
mrbean-bremen, line 1717: I think this check should be `#if PY_MAJOR_VERSION > 3 || (PY_MAJOR_VERSION == 3 && PY_MINOR_VERSION >= 4)`, otherwise it would be same as `#if PY_MAJOR_VERSION >= 3`.

mrbean-bremen (approved):

PR #49, opened 2021-06-23
Add isatty() method to PythonQtStdIn class

### LINKED ISSUE #48: Add .isatty() to PythonQtStdIn
Hello. I would like an `isatty` method in `stdin` as the one in `stdout`:

https://github.com/MeVisLab/pythonqt/blob/58ccd5393922d88fd320628aa2db787df1f1233b/src/PythonQtStdOut.cpp#L115-L117

May I open a PR to add it to https://github.com/MeVisLab/pythonqt/blob/master/src/PythonQtStdIn.cpp? This is to follow the advice I was given at https://github.com/Slicer/Slicer/issues/5705 to fix an issue using `pip` within the 3D Slicer Python console.

fepegar:
Resolves #48.

Context: https://github.com/Slicer/Slicer/issues/5705.

### DIFF
--- a/src/PythonQtStdIn.cpp
+++ b/src/PythonQtStdIn.cpp
@@ -62,9 +62,18 @@ static PyObject *PythonQtStdInRedirect_readline(PyObject * self, PyObject * args
   return Py_BuildValue("s", QStringToPythonConstCharPointer(string));
 }
 
+static PyObject *PythonQtStdInRedirect_isatty(PyObject * /*self*/, PyObject * /*args*/)
+{
+  Py_INCREF(Py_False);
+  return Py_False;
+}
+
 static PyMethodDef PythonQtStdInRedirect_methods[] = {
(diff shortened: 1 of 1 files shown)


lassoan (commented):
This looks good to me, but it would be better if you added a `bool _isatty{false}` member to `PythonQtStdInRedirect` and return the value of this in `isatty()`. This would allow using this class to stream input from an interactive terminal.

fepegar:
Thanks for reviewing, @lassoan.

> added a bool _isatty{false} member to PythonQtStdInRedirect

Sorry, I don't know how to do that. I don't know much C++.

lassoan:
@fepegar I could not update your pull request, so created a new one: https://github.com/MeVisLab/pythonqt/pull/50

I've tested it in Slicer and verified that `light-the-torch` package can be installed and imported correctly.

fepegar:
Great! Thanks, @lassoan. I see what you mean by adding a new member. It wasn't that hard! Next time I'll know what to do.

PR #50, opened 2021-06-23
Add isatty() method to PythonQtStdIn class

### LINKED ISSUE #48: Add .isatty() to PythonQtStdIn

lassoan:
pip and some other Python modules use this attribute to check if they are running in an interactive console.

Co-authored-by: Fernando Perez-Garcia <fernando.perezgarcia.17@ucl.ac.uk>

fixes #48

### DIFF
--- a/src/PythonQtStdIn.cpp
+++ b/src/PythonQtStdIn.cpp
@@ -48,6 +48,7 @@ static PyObject *PythonQtStdInRedirect_new(PyTypeObject *type, PyObject * /*args
   self = (PythonQtStdInRedirect *)type->tp_alloc(type, 0);
   self->_cb = NULL;
   self->_callData = NULL;
+  self->_isatty = false;
 
   return (PyObject *)self;
 }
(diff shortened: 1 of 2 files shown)


jcfr:
Thanks @lassoan 

Could you update the commit to include the following:

```
Co-authored-by: Fernando Perez-Garcia <fernando.perezgarcia.17@ucl.ac.uk>
```

For reference, here is a similar commit we contributed back almost 10 yrs ago ... https://github.com/commontk/PythonQt/commit/7132dba93064c2a02591b42305fecdd5d59702d3

lassoan:
> Could you update the commit to include the following:

This was already in the commit comment.

jcfr:
> This was already in the commit comment.

Issue is that the `Co-authored-by: ` is a "commit trailer" and should be at the end of the commit message otherwise it is ignored.

```diff
pip and some other Python modules use this attribute to check if they are running in an interactive console.
+
+ fixes #48
+
Co-authored-by: Fernando Perez-Garcia <fernando.perezgarcia.17@ucl.ac.uk>
-
- fixes #48
```

lassoan:
I did not know that the order matter. I've moved the coauthor message to the end.

PR #66, opened 2021-11-12
Queued slot calls

githubuser0xFFFF:
The merge request is a proposal for the implementation of queued slot calls.

In our application we run the Python interpreter in a worker thread and not in the UI main thread. That means, with the current implementation it is not possible/allowed to call slots of UI objects / widgets from our python scripts.

We changed the `PythonQtSlotInfo::invokeQtMethod` to check if the given object thread is the current thread. If the threads are equal, then we use the default implementation. If the slots are differents (Python interpreter in worker thread and Widget in main UI thread), then we call the slot via a blocking queued connection - like it is done from Qt in the `QMetaMehod::invoke` implementation.

In the proposal we provided two different implementations. The first one posts a `QMetaCallEvent` like the `QMetaMehod::invoke`  implementation. The advantages are, that the slot is called in the given object thread and that it has less overhead then the second implementation. The disadvandtage is, that is uses private API - the current implementation my only compile with Qt 5.15.

The second implementation uses a `PythonQtMetaCallDelegate` object that is created in the UI main thread. That means, the slot call is delegated to the UI main thread via invokation of a `PythonQtMetaCallDelegate` slot. This implementation does not use private API but the slots are always called in the UI main thread and not in the thread of the given object.

### DIFF
--- a/src/PythonQtMethodInfo.cpp
+++ b/src/PythonQtMethodInfo.cpp
@@ -654,3 +654,8 @@ void PythonQtSlotInfo::setGlobalShouldAllowThreads(bool flag)
 {
   _globalShouldAllowThreads = flag;
 }
+
+bool PythonQtSlotInfo::getGlobalShouldAllowThreads()
+{
+	return _globalShouldAllowThreads;
+}
--- a/src/PythonQtMethodInfo.cpp
+++ b/src/PythonQtMethodInfo.cpp
@@ -41,7 +41,11 @@
 
 #include "PythonQtMethodInfo.h"
 #include "PythonQtClassInfo.h"
+#include <QThread>
+#include <QSemaphore>
+#include <QCoreApplication>
 #include <iostream>
+#include <private/qobject_p.h>
 
 QHash<QByteArray, PythonQtMethodInfo*> PythonQtMethodInfo::_cachedSignatures;
 QHash<int, PythonQtMethodInfo::ParameterInfo> PythonQtMethodInfo::_cachedParameterInfos;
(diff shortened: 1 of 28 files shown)


githubuser0xFFFF:
I updated the merge request. I decided to use the 'PythonQtMetaCallDelegate' implementation to get rid of private API dependency. I also implemented proper exception handling for delegated meta calls.

With this implementation it is possible to run the python interpreter in a worker thread and call slots of widgets in the main UI thread.

mrbean-bremen (approved):

src/PythonQtMethodInfo.cpp
```diff
@@ -659,3 +652,78 @@ bool PythonQtSlotInfo::getGlobalShouldAllowThreads()
...
+/**
+ * If the current thread is not the main UI thread, then the metacall is
+ * delegated to the MetaCallDelegate. This helper object has been created in
+ * the main UI thread and will execute the metacall in the main UI thread.
+ */
+void PythonQtSlotInfo::invokeQtMethod(QObject* obj, PythonQtSlotInfo* slot, void** args)
+{
+  if (QThread::currentThread() != MetaCallDelegate.thread())
+  {
+    delegateMetaCall(obj, slot, args);
```
usiems, line 692: This assumes that this will only ever be done on objects in the main thread, without a check that the targeted object lives in the main thread. This could break (or at least slow down) a lot of code.

usiems:
No, this will not do. This code assumes that every Qt object lives in the main thread, which is an invalid assumption. One can start new Qt threads with their own event loops from Python (we do this in MeVisLab to handle asyncio tasks).

I also don't like that this will do cross-thread communication implicitly.

Personally I would prefer an approach where the QMetaObject wrapper gets a static invokeMethod method that takes the target object, the method name, connection style, and a list of arguments (almost like in C++, but more user-friendly). Of course this would mean figuring out which slot to use specifically (in case of overloads), but this is not much different from what is done already. Depending on whether Qt.QueuedConnection or Qt.BlockedQueuedConnection is used one would get a return value or not - if no return value is required this can be infinitely faster if the receiving thread is not idle in its event loop at the moment.

usiems (commented):
Some other complaints.

PR #72, opened 2022-06-24
ENH: checks to prevent infinite recursion

fbordignon:
I was discussing this issue on a fork, please check: https://github.com/commontk/PythonQt/pull/81

Basically I've integrated pyside2 with PythonQt by preventing a stack overflow that was happening because of an infinite recursion.
See results on: https://discourse.slicer.org/t/pythonqt-properties-shadowing-methods/16992


florianlink:
I wonder why that class has itself as its parent, shouldn’t that be fixed? PythonQt traverses the Qt metaobject parent hierarchy  and it should not be possible to have a class that derived  from itself. I suggest to find the reason for the problem and not fixing it at this place. Maybe you can fix it also by checking the parent info when it gets created, and not add itself as parent class in the first place.

fbordignon:
I see, I am wrapping a PythonQt qt layout with shiboken2 and adding a pyside2 widget to it:

```python
lw = PlotLayout.register().layout()
PlotLayout.show()

# wrap pythonQt instance with shiboken2 (pyside2)
pysidelayout = shiboken2.wrapInstance(hash(lw), QVBoxLayout)

pg.mkQApp()

# start of pyqtgraph plotting example https://github.com/pyqtgraph/pyqtgraph/blob/master/examples/Plotting.py
win = pg.GraphicsLayoutWidget(show=True, title="Basic plotting examples")
pysidelayout.addWidget(win)
```

So something assigns the parent as itself when I add the widget to the PythonQt layout. I don't know how that happens and don't have much time to look into it, so I am sorry but it is too much of a task for me right now. If you have any idea on how to find when this is happening, please let me know. Thanks!

jbowler:
This seems to be an assert but because the condition (class with itself as an ancestor) may perhaps be produced by code outside PythonQt a qFatal is probably a better bet; it does the same as an assert but with a potentially more helpful error message:

if (<class has itself as a parent>) qFatal("'%s' is malformed", <name of class>);

Issue #33, opened 2021-01-19
Segfault in PythonQtPrivate::createNewPythonQtInstanceWrapper


he-hesce:
Dear PythonQt community,

I recent upgraded from the old trunk r502 (sourceforge) to latest code on GitHub master branch, committish 9ea0fcf of 2 December 2020. I did this in order to get compatibility with Python 3.8. Using Python 2.7 and 3.6 have both worked fine with "r502".

Now when I try to use the latest PythonQt with code which previously worked fine, I get a segmentation fault. Looking at some (somewhat) recent commits I see there has been some work going on in the area where the segfault happens. Though it's hard for me to identify which commit introduced the issue.

The backtrace is here:

```
#0  0x00007ffff40f403f in PyObject_Call () at /lib64/libpython2.7.so.1.0
#1  0x00007ffff3d11dc0 in PythonQtPrivate::createNewPythonQtInstanceWrapper(QObject*, PythonQtClassInfo*, void*) ()
    at /home/xxx/git/qtbuild/libPythonQt-Qt5-Python2.7.so.3
#2  0x00007ffff3d13b11 in PythonQtPrivate::wrapPtr(void*, QByteArray const&, bool) () at /home/xxx/git/qtbuild/libPythonQt-Qt5-Python2.7.so.3
#3  0x00007ffff3db78c7 in PythonQtConv::createCopyFromMetaType(int, void const*) () at /home/xxx/git/qtbuild/libPythonQt-Qt5-Python2.7.so.3
#4  0x00007ffff3db8663 in PythonQtConv::QVariantListToPyObject(QList<QVariant> const&) () at /home/xxx/git/qtbuild/libPythonQt-Qt5-Python2.7.so.3
#5  0x00007ffff3dbc6b2 in PythonQtSignalTarget::call(_object*, PythonQtMethodInfo const*, void**, bool) ()
    at /home/xxx/git/qtbuild/libPythonQt-Qt5-Python2.7.so.3
#6  0x00007ffff3dbc7eb in PythonQtSignalTarget::call(void**) const ()
    at /home/xxx/git/qtbuild/libPythonQt-Qt5-Python2.7.so.3
#7  0x00007ffff3dbd04a in PythonQtSignalReceiver::qt_metacall(QMetaObject::Call, int, void**) () at /home/xxx/git/qtbuild/libPythonQt-Qt5-Python2.7.so.3
[...]

PR #79, opened 2022-12-23
Merge changes mostly related to compiler warnings

mrbean-bremen:
- replace NULL/0 with nullptr
- remove unused variables
- add override where needed
- use C++ struct syntax
- fix some more warnings from newer compilers

### DIFF
--- a/src/PythonQtBoolResult.cpp
+++ b/src/PythonQtBoolResult.cpp
@@ -66,54 +66,54 @@ static int PythonQtBoolResult_nonzero(PyObject *obj)
 
 // we override nb_nonzero, so that one can do 'if' expressions to test for a NULL ptr
 static PyNumberMethods PythonQtBoolResult_as_number = {
-  0,      /* nb_add */
-  0,      /* nb_subtract */
-  0,      /* nb_multiply */
+  nullptr,      /* nb_add */
+  nullptr,      /* nb_subtract */
+  nullptr,      /* nb_multiply */
 #ifndef PY3K
-  0,      /* nb_divide */
+  nullptr,      /* nb_divide */
 #endif
-  0,      /* nb_remainder */
-  0,      /* nb_divmod */
-  0,      /* nb_power */
-  0,      /* nb_negative */
-  0,      /* nb_positive */
-  0,      /* nb_absolute */
+  nullptr,      /* nb_remainder */
+  nullptr,      /* nb_divmod */
+  nullptr,      /* nb_power */
+  nullptr,      /* nb_negative */
+  nullptr,      /* nb_positive */
+  nullptr,      /* nb_absolute */
   PythonQtBoolResult_nonzero,      /* nb_nonzero / nb_bool in Py3K */
-  0,      /* nb_invert */
-  0,      /* nb_lshift */
-  0,      /* nb_rshift */
-  0,    /* nb_and */
-  0,    /* nb_xor */
-  0,    /* nb_or */
-#ifndef PY3K
-  0,      /* nb_coerce */
+  nullptr,      /* nb_invert */
+  nullptr,      /* nb_lshift */
+  nullptr,      /* nb_rshift */
+  nullptr,      /* nb_and */
+  nullptr,      /* nb_xor */
+  nullptr,      /* nb_or */
+#ifndef PY3K   
+  nullptr,      /* nb_coerce */
 #endif
-  0,      /* nb_int */
-  0,      /* nb_long  / nb_reserved in Py3K */
-  0,      /* nb_float */
+  nullptr,      /* nb_int */
+  nullptr,      /* nb_long  / nb_reserved in Py3K */
+  nullptr,      /* nb_float */
 #ifndef PY3K
-  0,      /* nb_oct */
-  0,      /* nb_hex */
+  nullptr,      /* nb_oct */
+  nullptr,      /* nb_hex */
 #endif
-  0,      /* nb_inplace_add */
-  0,      /* nb_inplace_subtract */
-  0,      /* nb_inplace_multiply */
+  nullptr,      /* nb_inplace_add */
+  nullptr,      /* nb_inplace_subtract */
+  nullptr,      /* nb_inplace_multiply */
 #ifndef PY3K
-  0,      /* nb_inplace_divide */
+  nullptr,      /* nb_inplace_divide */
 #endif
-  0,      /* nb_inplace_remainder */
-  0,      /* nb_inplace_power */
-  0,      /* nb_inplace_lshift */
-  0,      /* nb_inplace_rshift */
-  0,      /* nb_inplace_and */
-  0,      /* nb_inplace_xor */
-  0,      /* nb_inplace_or */
-  0,      /* nb_floor_divide */
-  0,      /* nb_true_divide */
-  0,      /* nb_inplace_floor_divide */
-  0,      /* nb_inplace_true_divide */
+  nullptr,      /* nb_inplace_remainder */
+  nullptr,      /* nb_inplace_power */
+  nullptr,      /* nb_inplace_lshift */
+  nullptr,      /* nb_inplace_rshift */
+  nullptr,      /* nb_inplace_and */
+  nullptr,      /* nb_inplace_xor */
+  nullptr,      /* nb_inplace_or */
+  nullptr,      /* nb_floor_divide */
+  nullptr,      /* nb_true_divide */
+  nullptr,      /* nb_inplace_floor_divide */
+  nullptr,      /* nb_inplace_true_divide */
 #ifdef PY3K
-  0,      /* nb_index in Py3K */
+  nullptr,      /* nb_index in Py3K */
 #endif
 };
 
@@ -122,37 +122,37 @@ PyTypeObject PythonQtBoolResult_Type = {
     "BoolResult",
     sizeof(PythonQtBoolResultObject),
     0,
-    0,     /* tp_dealloc */
-    0,          /* tp_print */
-    0,          /* tp_getattr */
-    0,          /* tp_setattr */
-    0,
+    nullptr,      /* tp_dealloc */
+    0,            /* tp_vectorcall_offset */
+    nullptr,      /* tp_getattr */
+    nullptr,      /* tp_setattr */
+    nullptr,
     (reprfunc)PythonQtBoolResult_repr,      /* tp_repr */
     &PythonQtBoolResult_as_number,          /* tp_as_number */
-    0,          /* tp_as_sequence */
-    0,          /* tp_as_mapping */
-    0,      /* tp_hash */
-    0,      /* tp_call */
-    0,          /* tp_str */
-    0,    /* tp_getattro */
-    0,          /* tp_setattro */
-    0,          /* tp_as_buffer */
+    nullptr,      /* tp_as_sequence */
+    nullptr,      /* tp_as_mapping */
+    nullptr,      /* tp_hash */
+    nullptr,      /* tp_call */
+    nullptr,      /* tp_str */
+    nullptr,      /* tp_getattro */
+    nullptr,      /* tp_setattro */
+    nullptr,      /* tp_as_buffer */
     Py_TPFLAGS_DEFAULT,/* tp_flags */
-    "Result object that is useful for bool* arguments",          /* tp_doc */
-    0,    /* tp_traverse */
-    0,          /* tp_clear */
-    0,          /* tp_richcompare */
-    0,          /* tp_weaklistoffset */
-    0,          /* tp_iter */
-    0,          /* tp_iternext */
-    0,          /* tp_methods */
-    0,       /* tp_members */
-    0,       /* tp_getset */
-    0,          /* tp_base */
-    0,          /* tp_dict */
-    0,                         /* tp_descr_get */
-    0,                         /* tp_descr_set */
-    0,                         /* tp_dictoffset */
+    "Result object that is useful for bool* arguments",     /* tp_doc */
+    nullptr,      /* tp_traverse */
+    nullptr,      /* tp_clear */
+    nullptr,      /* tp_richcompare */
+    0,            /* tp_weaklistoffset */
+    nullptr,      /* tp_iter */
+    nullptr,      /* tp_iternext */
+    nullptr,      /* tp_methods */
+    nullptr,      /* tp_members */
+    nullptr,      /* tp_getset */
+    nullptr,      /* tp_base */
+    nullptr,      /* tp_dict */
+    nullptr,      /* tp_descr_get */
+    nullptr,      /* tp_descr_set */
+    0,            /* tp_dictoffset */
     (initproc)&PythonQtBoolResult_init,      /* tp_init */
 };
 
--- a/src/PythonQtClassWrapper.cpp
+++ b/src/PythonQtClassWrapper.cpp
@@ -51,34 +51,34 @@
 
 static PyObject* PythonQtInstanceWrapper_invert(PythonQtInstanceWrapper* wrapper)
 {
-  PyObject* result = NULL;
+  PyObject* result = nullptr;
   static QByteArray memberName = "__invert__";
   PythonQtMemberInfo opSlot = wrapper->classInfo()->member(memberName);
   if (opSlot._type == PythonQtMemberInfo::Slot) {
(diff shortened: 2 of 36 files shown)


he-hesce:
I think it would be better to use Q_NULLPTR as we are using Qt anyway so it is available as nullptr is C++11 and  PythonQt still works with pre-C++11 compilers like on RHEL 6. Q_NULLPTR is there as macro so support both pre and post C++11. On C++11 and later it is probably defined as nullptr.

mrbean-bremen:
Thanks - good point about compatibility! I actually wasn't aware that this still supports pre-C++11 compilers. Looking at the [supported platforms in Qt5](https://doc.qt.io/qt-5/supported-platforms.html) I see that they support gcc5 as a minimum, which does have C++11 support to my knowledge, so I'm not sure how much sense it makes to support older compilers in `PythonQt`.

On a related note - it would really help to have some CI builds with different compilers to catch these kinds of problems. I may have a look at this some time later. Also, switching to CMake probably also makes sense at some time, as Qt has done.

For some background: we have a private fork of `PythonQt` that has these (and other) changes, and our goal is to eventually make this obsolete and just use the published version, therefore I want to merge back as much as possible (and what makes sense).

he-hesce:
RHEL/CentOS 6 (still heavily used in scientific and production environments) ship with Qt 5 and gcc 4.4 (and Python 2.7). Most scientific Python code in the world is still on Python 2.6 and 2.7.

I never see a reason to break compatibility with pre-C++11 compilers just for the sake of it, specially when there are fully working and supported compatibility stuff like Q_NULLPTR. 

The only real nice thing in C++11 would be lambda functions which saves a lot of coding 
compared to using QSignalHandler but people who can require C++11 can then choose to use those in their code. But it is nice if you're supporting libraries such as PythonQt doesn't require C++11 unless it is an absolute necessity.

mrbean-bremen:
Well, I see your point, though I can't say that I agree with all of what you wrote here. There have been reasons for Python 3 as well as for newer C++ versions besides the hype of better and newer, and there are other relevant new things in C++ besides lambdas (though I agree about them). I have written K&R C code myself in my time, and I wouldn't want to write it in the same way today. Anyway, I don't think it will get us anywhere discussing the merits of new features in programming languages...

More to the point: I see no problem in using the compatibilty stuff if it makes sense. I also understand that there is a lot of old software out there that still needs to run, but my question is if they need new versions of PythonQt. As I understand it, PythonQt will mostly evolve to accomodate newer Qt and Python versions, not to add more features for the existing ones. Also, there is always the possibility to make patch releases for an older version, if really needed. 

What I would like to see is a clearly defined minimum compiler version for a specific version of PythonQt, and a CI build that tests this version with several compilers. I really don't think that we will need builds for pre-C++11 compilers for a Qt6 version, but I may be wrong of course. I'll probably wait and see if somebody else chimes in...

he-hesce:
It is true one can always freeze on an older version of PythonQt and backport fixes from newer versions if needed. I just thought I would voice my opinion as I always run into code changing just because its is cool to use the latest features when the old code worked just fine and then I end up maintaining patch sets to change the code back for compatibility reasons. Cheers!

mrbean-bremen:
Yes, supporting older code bases is certainly a common problem. I shudder if I imagine the plethora of old Fortran 66 and Cobol code still out there...
I think for small projects like this one the best option is to to use release branches to be able to backport fixes if needed, but otherwise just go ahead. I have done this with Python projects with a Python 2 - supporting release branch, and while I did backport bugfixes a couple of times when asked to, that was far less effort than having to support Python 2 all along in the current code. 
I just made a new GitHub release for the current state, and if needed we can also make a branch from the release tag. As I wrote, I think it makes sense to support at most the same compilers as Qt does, e.g. gcc5 for Qt 5.15 and gcc 9 for Qt 6.2.

he-hesce:
I still maintain Fortran 77 code :-) Nobody is going to rewrite it when it works. And it continues to link with and be used from C99 and C++ code.

I already have to use the official revert-refcount-fixes branch as otherwise my use of PythonQt crashes (see issue #33). So a branch compatible with pre-C++11 is okay.

As for Python 2/3, it's not that hard to write code which works on both 2.7 and 3.6/8/9. Just takes some effort and testing. Thought when done better than implementing new features in two copies of the code.

It should be noted that RHEL/CentOS 7.9 ships with Qt 5.9 and gcc 4.8 and it is not EOL yet (1.5 years left). So not everyone is using Qt 5.15. Qt 6 will likely be something production environments can look at in 10-15 years from now. 5.x series will be used for a long time.

You may find https://www.nature.com/articles/d41586-021-01431-y a fun read.

Cheers and Happy New Year!

mrbean-bremen:
> I still maintain Fortran 77 code :-) Nobody is going to rewrite it when it works. And it continues to link with and be used from C99 and C++ code.

Well, lucky you :) I didn't have much to do with Fortran - though when I did (also Fortran 77), I can't say that I liked it (compared to K&R C I was mainly using at the time). 

> As for Python 2/3, it's not that hard to write code which works on both 2.7 and 3.6/8/9. Just takes some effort and testing. Though when done better than implementing new features in two copies of the code.

Yes, I did this also, but it really depends on the usage. If there is almost no use case for Python 2 left, and some Python3 features are actually needed for full functionality, it just isn't worth it to maintain both versions. But I know that are still many packages out there that only support Python 2, though with the pPython 2 sunsetting[(https://python3statement.org/) effort their number has been largely reduced.

> so not everyone is using Qt 5.15. Qt 6 will likely be something production environments can look at in 10-15 years from now. 5.x series will be used for a long time. 

Well, that's a bit overly pessimistic IMHO. I know that this is certainly true for some environments, but I also know that production environments have switched to Qt 6, or switching right now. But I get your point, especially for embedded and high-risk environments.

> You may find https://www.nature.com/articles/d41586-021-01431-y a fun read.

Thanks, always one for a fun read :)

Happy New Year to you too!

he-hesce:
If you just moved to RHEL/Rocky 9 you will be using Qt 5.15 for another 10-15 years supported and probably more than that unsupported. Perhaps RHEL/Rocky 10 will ship with Qt 6 in 3-4 years from now.

mrbean-bremen:
> As for Python 2, RHEL/CentOS 7 still ships with and supports Python 2.7 for another 1.5-6.5 years from now. Hence, if you're on RHEL/CentOS 7 that is what you have to use. They threw in Python 3.6 in the final feature update to 7 but without any modules and not default, so half-hearted desperate move before 7 moved into security-only support mode.

Yes, I know, but that just means that they will use a version of the package that still supports that version, and not the newest one. CentOS is basically dead (thanks to Red Hat), e.g. while it still may be used for a long time, it is not expected to need any new features. They won't get any more security fixes for Python 2 anyway, so they should not expect such fixes for any other Python 2 packages (though if really needed, these can still be done).

he-hesce:
All true what you say. Though the fun for us developers begins when you have to support the latest version of your own software (containing the feature the users asked for) on RHEL 6, 7, 8 & 9 as your users can be on any of these platforms and unable to move beyond what they are on. That is when you appreciate your dependencies such as PythonQt still working on all of these with the same code so you don't need to keep different copies of PythonQt in your source tree for linking into application depending on gcc/Python version. Of course, I know this is a lot to ask for an open-source project to support when these ancient versions of gcc/Python are officially not supported anymore. So all I am asking for in this pull request is to use Q_NULLPTR instead of nullptr as an example because that will give you the same benefit on gcc 4.8+ as using nullptr directly but still continue to work for those without nullptr.

mrbean-bremen:
Well, that would be a fair point, but while trying to setup building a CI build for PythonQt (I will make a separate PR for this) I noticed that the current version already does not compile with pre-C++11 compilers. 
The [gcc 5.5 build failed](https://github.com/mrbean-bremen/pythonqt/actions/runs/3800253038/jobs/6463552588) because of that (I didn't add the `-std=c++11` flag that is needed here to compile C++11 code, so it compiled in c++98 mode).
It fails on some move constructors and move assignment operators that are in the code since 2018, so I really don't see how you can use the current version with a pre-gcc4.8 compiler. Maybe I'm missing something here...

he-hesce:
There are indeed pre-existing code which has to be re-written back to C++98 each time one wants to use a new release. I would have stayed at r502 if it wasn't for that I needed Python 3.8+ support. And then when moving up to that commit I have to rebase to remove a bunch of commits to avoid segfaults (#33). So it ain't just downloading the HEAD of mater branch and get going :-)

I also ran into a header file having been moved in Python 3.10 on Ubuntu 22.04 so I will submit an issue for that.

florianlink:
I think we should revert the leak fixes that have led to the segfaults. It
seems like one if the changes was too aggressive, and since we did not spot
which one, I would vote for reverting.

mrbean-bremen:
@he-hesce - is it possible to reproduce this segfault, maybe in a docker container that can be run in the CI? This would make it easier to pin it down to the problematic change. 

Using the Qt macros instead of the C++11 constructs should not be a problem, but this still leaves at least the move constructors and assignment operators - I didn't check if there are more problems with compiling in C++-98 mode.  Though maybe that isn't actually much and can actually be wrapped...

he-hesce:
The problem reproducing is in creating a minimal test case for this to reproduced. Right now there is a whole application with many layers where this is happening which also makes it difficult to create a test case as it requires the whole call chain and signaling and so on. We discussed this in #33. Another user also had the same issue. Basically it comes down to some race condition between calls and too aggressive garbage collection where it seems objects are freed while they are still in use.

mrbean-bremen:
Ok, I had to ask :) In this case it is probably really the best to just revert the changes as in that branch, as proposed by @florianlink. Though I would at least wait for the opinion of @usiems, who created the `revert-refcount-fixes` branch.

he-hesce:
Here are two changes from C++11 to Qt syntactic sugar macros to libsrc/PythonQt/src/PythonQtMethodInfo.cpp:

```
-  for (const ParameterInfo& info : _parameters) {
+  Q_FOREACH (const ParameterInfo& info, _parameters) {
```
```
-  for(const PythonQtSignalTarget& t : _targets) {
+  Q_FOREACH (const PythonQtSignalTarget& t, _targets) {
```

mrbean-bremen:
> Q_FOREACH used in many other places before this but then new code was added using C++11-style for (C++11 range-based for loops) instead.

This is a case which I don't really like - Q_FOREACH is convenient, but implemented as an ugly macro that is problematic to debug. Well, you can't have everything...

he-hesce:
I think **Q_FOREACH** is Qt 4 and Qt 5 got **foreach**. Not sure if they are different or the same implementation behind the scenes.
https://doc.qt.io/qt-6/foreach-keyword.html

Though this is not a hill I am going to die on. Just voicing my opinion that compatible code is preferable if possible.

mrbean-bremen:
Yeah, as far as I remember `foreach` is the same as `Q_FOREACH`, just better readable. As they write on the the site you point to:
> Note: The foreach keyword was introduced before the C++11 range-based loops existed. New code should prefer C++11 range-based loops.

mrbean-bremen:
I tried compiling with -std=c++98 and noticed that this is not possible with Qt 5.15 - it does use C++11 itself (somewhat predictable given the supported compiler versions). There are many things that cannot easily be hidden just using the provided macros.

As further changes in PythonQt will be mainly for Qt 5.15 and Qt 6 support, I don't think it is worth the effort to try to make the master branch C++98-compatible. We probably should make sure that it is C++11-compatible by using the respective compiler flag for Qt 5 builds, though.

he-hesce:
Ok. C++11 is fine :-) RHEL/CentOS 7 has gcc 4.8 which supports C++11. However, RHEL 7 still ships with Qt 5.9.7 so not breaking compatibility with pre-Qt-5.15 would be nice.

Luckily RHEL/Rocky 8 and RHEL/Rocky 9. both ship with Qt 5.15.3.

he-hesce:
I am still using generated_cpp_56 as it is the last one including QtWebKit which is still shipped with RHE: 67/8/9 and used by many, though not part of official Qt anymore. It is maintained/developed separately from the Qt Project now. I once managed to re-generate the files for QtWebKit for Qt 5.9 but could never replicate it for Qt 5.12 and 5.15. SO still using good old stable generated_cpp_56 which works fine as long I don't need to use any Qt functionality added after 5.6 (which is actually never).

mrbean-bremen:
> RHEL/CentOS 7 has gcc 4.8 which supports C++11. However, RHEL/CentOS 7 still ships with Qt 5.9.7 so not breaking compatibility with pre-Qt-5.15 would be nice.

A docker image with CentOS 7 that builds qtmake could ensure this, I will try to set this up.

mrbean-bremen:
This is ready for merge - it basically merges all relevant changes that we had in our local fork (except the CMAKE-specific files). I split the changes into several commits, and I added the define `PYTHON2` for a Python3-specifix feature (last commit) made by Florian. If compiling for Python 2, this has currently to be set manually as done in the CentOS build (I failed to correctly define it automatically in the prf).

src/PythonQtQFileImporter.h
```diff
@@ -43,22 +43,22 @@
 //----------------------------------------------------------------------------------
 
 #include "PythonQtPythonInclude.h"
-#include "PythonQtImportFileInterface.h"
```
usiems, line 46: I think the quotes were correct.

src/PythonQtClassWrapper.cpp
```diff
@@ -593,49 +593,49 @@ static PyObject * PythonQtClassWrapper_repr(PyObject * obj)
...
+    nullptr,                         /* tp_richcompare */
+    0,                               /* tp_weaklistoffset */
+    nullptr,                         /* tp_iter */
+    nullptr,                         /* tp_iternext */
     #ifdef PY3K
-    PythonQtClassWrapper_methods,                   /* tp_methods */
+    PythonQtClassWrapper_methods,    /* tp_methods */
     #else
-    0,                   /* tp_methods */
+    0,                               /* tp_methods */
```
usiems, line 626: I think this should be nullptr too.

src/PythonQtBoolResult.cpp
```diff
@@ -66,54 +66,54 @@ static int PythonQtBoolResult_nonzero(PyObject *obj)
...
-#ifndef PY3K
-  0,      /* nb_coerce */
+  nullptr,      /* nb_invert */
+  nullptr,      /* nb_lshift */
+  nullptr,      /* nb_rshift */
+  nullptr,      /* nb_and */
+  nullptr,      /* nb_xor */
+  nullptr,      /* nb_or */
+#ifndef PY3K   
+  0,            /* nb_coerce */
```
usiems, line 89: I think this should be nullptr too.

src/PythonQtBoolResult.cpp
```diff
@@ -66,54 +66,54 @@ static int PythonQtBoolResult_nonzero(PyObject *obj)
...
-  0,      /* nb_int */
-  0,      /* nb_long  / nb_reserved in Py3K */
-  0,      /* nb_float */
+  nullptr,      /* nb_int */
+  nullptr,      /* nb_long  / nb_reserved in Py3K */
+  nullptr,      /* nb_float */
 #ifndef PY3K
-  0,      /* nb_oct */
-  0,      /* nb_hex */
+  0,            /* nb_oct */
```
usiems, line 95: I think this should be nullptr too.

src/PythonQtBoolResult.cpp
```diff
@@ -66,54 +66,54 @@ static int PythonQtBoolResult_nonzero(PyObject *obj)
...
 #endif
-  0,      /* nb_inplace_add */
-  0,      /* nb_inplace_subtract */
-  0,      /* nb_inplace_multiply */
+  nullptr,      /* nb_inplace_add */
+  nullptr,      /* nb_inplace_subtract */
+  nullptr,      /* nb_inplace_multiply */
 #ifndef PY3K
-  0,      /* nb_inplace_divide */
+  0,            /* nb_inplace_divide */
```
usiems, line 102: I think this should be nullptr too.

src/PythonQtThreadSupport.h
```diff
@@ -40,8 +40,8 @@
 //----------------------------------------------------------------------------------
 
 
-#include "PythonQtPythonInclude.h"
```
usiems, line 43: I think the quotes were correct here.

usiems (approved):

PR #80, opened 2022-12-29
Set up CI builds for different gcc versions

mrbean-bremen:
This is a first attempt to setup some CI builds, currently only building PythonQt using qmake/make using gcc 7, 9 and 11.
This just uses the GCC docker images and the Python and Qt images that are installed there (Python 3.7.3 / Qt 5.11.3 for gcc 7 and 9, Python 3.9.2 / Qt 5.15.2 for gcc 11).

Probable tasks for future PRs:
- setup Windows and Mac builds
- use specific Qt and Python version combinations
- run the created generator for a specific Qt version and provide the created code
- a bit unrelated - switch to CMake (there are already several forks who have done this)

I also tried to use gcc 5 and 6, but if just using the standard GCC docker images with the standard Python and Qt packages, they run with Python 3.4 and Qt 5.3.2, which is not supported with the current version (does not compile due to missing member functions).

### DIFF
--- a/.github/workflows/Dockerfile_gcc
+++ b/.github/workflows/Dockerfile_gcc
@@ -1,0 +1,38 @@
+FROM gcc:#GCC#
+
+RUN apt-get update && apt-get install -y --force-yes \
+    unzip \
+    wget \
+    python3 \
+    python3-dev \
+    libpython3-dev \
+    qtbase5-dev \
+    qtbase5-private-dev \
+    qtchooser \
+    qt5-qmake \
+    qtbase5-dev-tools \
+    qttools5-dev \
+    qtdeclarative5-dev \
+    libqt5svg5* \
+    libqt5xmlpatterns5* \
+    libqt5multimedia5* \
+    libqt5multimediawidgets5* \
+    libqt5qml5* \
+    libqt5quickwidgets5* \
+    qtmultimedia5-dev
+RUN apt-get clean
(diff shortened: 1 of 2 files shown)


he-hesce:
Not sure if this is possible but keeping the qmake build system and adding cmake as an option would be nice. As I build and link PythonQt in our source tree where we already use qmake, building of PythonQt is as easy as just adding it to SUBDIRS in the .pro file in the subdirectory above. If PythonQt is switched to cmake, I would need to set up some other procedure to build and link it with existing qmake build system I think. As qmake is working great I don't have any real reason to transition to cmake which from experience using it seems more complicated than qmake. Though I can understand people using cmake in their project would like to easily include PythonQt directly the same way I do with qmake.

mrbean-bremen:
It certainly shall be possible to leave in the qmake files, though in the long run it is probably not feasible to maintain two build systems. For Qt 5 this should certainly not be a problem.

mrbean-bremen:
Thinking about this again, we actually may not need builds with different gcc versions - it should be sufficient to just use a current version with the lowest needed compatibility setting (e.g. `-std= c++98` if we still want to support this), and use images with different Python and Qt versions instead. This PR can just serve as a startpoint...

mrbean-bremen:
I'm going to merge this to have a start point for future changes (can't break anything...)

PR #84, opened 2022-12-31
Add CentOS7 docker image to CI

mrbean-bremen:
- uses gcc 4.8.5, Python 3.6.8 and Qt 5.9.7
- also remove gcc 9 build (keep gcc 7 and 11)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,7 +6,10 @@ on:
       - master
   pull_request:
       
-   
+defaults:
+  run:
+    shell: bash
+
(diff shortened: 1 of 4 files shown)


he-hesce:
Do note that Python 2.7 is still the default and supported Python version on RHEL/CentOS 7. 3.6.8 is only provided as last ditch effort.

mrbean-bremen:
I see... I will check this some time later.

mrbean-bremen:
@he-hesce - changed to Python 2 now.

PR #86, opened 2023-01-04
Add wrapper generation to CI builds

mrbean-bremen:
No checks are currently made that the wrappers are generated correctly, but the generator output can be checked for problems. Note that the generator for 5.15 still has a lot of warnings and generates less wrappers compared to 5.11 (running on gcc7).
For comparison the number of wrappers generated as logged by the generator:

|Qt version|Build on|Nr wrappers|
|:-----------:|:-----------:|:-----------:|
|5.9|CentOS image|723|
|5.11|GCC 7 image|820|
|5.12|local build|836|
|5.15|GCC 11 image|698|

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         runs-on: [ubuntu-latest]
-        docker-image: [gcc7, gcc11, centos7]
+        docker-image: [centos7, ubuntu2004, ubuntu2204]
 
     steps:
     - uses: actions/checkout@v3
--- a/.github/workflows/dockerfiles/Dockerfile_centos7
+++ b/.github/workflows/dockerfiles/Dockerfile_centos7
@@ -25,8 +25,21 @@ WORKDIR work
 ARG QT_SELECT=qt5
 RUN uname -a; gcc --version | grep "gcc"; python --version; qmake-qt5 --version
 
+CMD ["make"]
+
+# 1. build PythonQt and the generator
+
 RUN qmake-qt5 -r PythonQt.pro \
     PYTHON_VERSION=$(python --version | cut -d " " -f 2 | cut -d "." -f1,2) \
     PYTHON_DIR=$(which python | xargs dirname | xargs dirname)
+RUN make
 
-CMD ["make"]
+# 2. generate wrappers using the built generator
+
+# workaround to allow to find the Qt include dirs for installed standard qt packages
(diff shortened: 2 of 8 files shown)


he-hesce:
It should be noted that Ubuntu 20.04 LTS (supported until April 2025) uses at 5.12. So ensuring PythonQt works with 5.9, 5.12, 5.15 would be nice.

Also, CentOS 7/8/9 and Ubuntu 18.04/20.04/22.04 offers QtWebKit as supported package even though Qt company no longer supports it. QtWebKit is developed and supported by the old developer of it and open-source team. It would be cool if the pre-generated headers shipping with PythonQt did include the QtWebKit interface. This could be done by ensuring QtWebKit package is installed on the corresponding platform before generating the interface file.

he-hesce:
Common long-term support production platform in active support window:

- RHEL/CentOS 7, Qt 5.9, Python 2.7, gcc 4.8
- RHEL/Rocky 8, Qt 5.15, Python 3.6, gcc 8.5
- RHEL/Rocky 9, Qt 5.15, Python 3.9, gcc 11.3
- Ubuntu 20.04 LTS, Qt 5.12, Python 3.8, gcc 9.4
- Ubuntu 22.04 LTS, Qt 5.15, Python 3.10, gcc 11.3

he-hesce:
As for qmake, referenced in issue linked above, it would indeed be great as you say to keep qmake support for Qt 5 as I am sure many projects using Qt 5 will not switch build system when nothing is broken (and changing build system is major work). Qt 5.15 will continue to be used for at least another decade or longer given its active support window on production/enterprise platforms.

mrbean-bremen:
I was going to replace the GCC images with standard Ubuntu images anyway, so this will replace Qt 5.11 with Qt 5.12.
Leaving qmake support for Qt 5 is just a matter of not removing the pri/pro/prf files, so this should not be a big deal.

As for QtWebKit - you could make a PR that adds it at least to the CI images - you probably know how to install it under CentOS and Ubuntu.

he-hesce:
I could clone the latest code and look into a full PR. However, it would be a one-liner installing the package. The package is qt5-qtwebkit-devel on RHEL/CentOS/Rocky and libqt5webkit5-dev on Ubuntu.

So

```
yum install qt5-qtwebkit-devel
```

on RHEL/CentOS/Rocky, and

```
apt install libqt5webkit5-dev
```

on Ubuntu.

mrbean-bremen:
Ah right, of course - I had somehow missed that this is the same issue... Thanks!

PR #91, opened 2023-01-31
Fix build with Python 3.11

iakov:
Python 3.11 removed some fields in the PyCodeObject, [but introduced access functions](https://docs.python.org/3/c-api/code.html#c.PyCode_GetVarnames). Thus, the similarly named function is enough to fix the build issue.


mrbean-bremen (approved):
Thanks!

PR #92, opened 2023-02-01
Fix install paths by adding INSTALL_PREFIX

yurivict:
Otherwise install and lib folders are installed into root.

### DIFF
--- a/extensions/PythonQt_QtAll/PythonQt_QtAll.pro
+++ b/extensions/PythonQt_QtAll/PythonQt_QtAll.pro
@@ -56,11 +56,11 @@ unix {
   QMAKE_PKGCONFIG_VERSION = $$VERSION
 }
 
-unix: target.path = /lib
+unix: target.path = $${INSTALL_PREFIX}/lib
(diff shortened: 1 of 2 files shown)


extensions/PythonQt_QtAll/PythonQt_QtAll.pro
```diff
@@ -56,11 +56,11 @@ unix {
   QMAKE_PKGCONFIG_VERSION = $$VERSION
 }
 
-unix: target.path = /lib
+unix: target.path = $${INSTALL_PREFIX}/lib
 win32: target.path = /
 
 headers.files = $${HEADERS}
-headers.path = /include
+headers.path = $${INSTALL_PREFIX}//include
```
mrbean-bremen, line 63: There seems to be an unwanted double slash

yurivict:
I corrected the double slash.

mrbean-bremen (approved):
Thanks - looks good to me! I will wait a bit with merging in case someone else will chime in...

Issue #13, opened 2020-06-18
About Qt6


BlueGhostStudio:
Qt6 is coming. Are there any plans to update PythonQt to support qt6?

thanks.

florianlink:
No, we have not looked at Qt 6 yet. But I guess basic features like the moc
and signals slots will stay, so it should not be hard to keep PythonQt
working. The wrapper generator is another thing, it might be needed to be
rewritten, e.g. using PySide shiboken technology.

SeekAWayOut:
Thanks for your respond

jamesobutler:
@florianlink 🙏 Thank you for supporting this project. Would you be willing to check-in generated wrappers for Qt 5.15? It is the last minor version of Qt 5 before Qt 6.  This would be helpful as a final point before going to Qt 6 for this project.

Murmele:
Thank you @florianlink  for that answer. I'm thinking about using pythonqt for another project, but was not shure about the wrapper generator because the last official commit at qt side was 3 years ago.

Murmele:
I'm interested in helping port to PySide for Qt6. @florianlink Do there exist a communication channel if I have some questions regarding the code?

florianlink:
No, I am probably the only one you can ask... florianlink at google mail

davidkhess:
@florianlink is there any word on Qt 5.15 generated wrappers being added? I'm not quite sure how to invoke `python_generator` but I think I'm doing it right with:

```
>set QTDIR=C:\Qt\5.15.2\msvc2019_64
>pythonqt_generator.exe
```

It throws a lot of warnings and then a lot of classes are missing in the result. 

For instance, QtGui builtin is missing QFont, QMatrix, QPen, and more.

I'm not sure if there is something wrong with my environment or if the typesystem_*.xml files need updating. I've turned up the debugging but it doesn't provide any clues.

Is there any advice you can provide on how to fix this?

davidkhess:
I should add, I'm able to successfully build and use the 5.11 generated wrapper with 5.15.2. I just need to access a new method on QJSEngine that was added in 5.12.

florianlink:
Probably some newer C++ features that  the generator parser stumbles
over... I did not yet look into it.

jamesobutler:
@florianlink Having been 1 year since the last comment on this thread, is PythonQt still an active project with plans for Qt6 support? Or are developers encouraged to use [Pyside6](https://pypi.org/project/PySide6/) instead?

[...]

PR #93, opened 2023-02-06
CI for Windows

iakov:
* Initial CI for Windows via GHA with `make check` to run unit-tests
* Basic `clang-msvc` support
* Fixes for `msvc` and `mingw`
* Support for LTO (`CONFIG+=ltcg`) builds on all platforms
* Minor code improvements, including build scripts unification

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -173,7 +173,7 @@ jobs:
         set -eu
         PYTHON_VERSION_FULL=$(python --version 2>&1 | cut -f 2 -d ' ')
         PYTHON_VERSION_SHORT=$(cut -f 1,2 -d . <<< $PYTHON_VERSION_FULL)
-        QT_VERSION_FULL=$($Qt5_DIR/bin/qmake -query QT_VERSION)
+        QT_VERSION_FULL=$($Qt5_Dir/bin/qmake -query QT_VERSION)
         QT_VERSION_SHORT=$(cut -f 1,2 -d . <<< $QT_VERSION_FULL)
         MACOS_VERSION_FULL=$(sw_vers -productVersion)
         MACOS_VERSION_SHORT=$(cut -f 1,2 -d . <<< $MACOS_VERSION_FULL)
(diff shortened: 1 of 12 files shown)


iakov:
More fixes from my repo. This is a small step towards Windows CI, that I already have done. But since there are lots of changes, I'll try to split to a few PRs.

iakov:
I was able to move from my repo the essential part for Win CI without other significant changes, so PR was renamed.

iakov:
Please, re-run the failed Ubuntu workflow. Maybe servers a back online.

mrbean-bremen (approved):
Nice work, thanks a lot! 
As usually, I will wait a bit before merging, to give others a chance to  have a look.

iakov:
Just FYI: It is very dangerous to use MSVC-built Python with  mingwXXX Qt runtime because of C runtime incompatibilities. I've encountered problems, but simple fix saved me from crash. The problem in my case was in `FILE*` data structure. Also, compiler incompatibilities can cause even Python's internal data structures to have different representations in pre-compiled `pythonXY.dll` and `PythonQt-Qt5-PythonX.Y.dll`. Please, take this into account. Probably, the pure MSYS build should be added to CI and some checks and warnings should be added to `python.prf`. If this makes sense, an issue may be opened for this case.

mrbean-bremen:
>  If this makes sense, an issue may be opened for this case.

It totally makes sense in my opinion - feel free to open an issue!

jamesobutler:
Hi all! I’ve observed the recent development here in the PythonQt repo recently which is great to see 🙏🏻. Are you all associated with MeVisLab? Are you all prepping the testing infrastructure and modernizing the build infrastructure with the specific next step of updating to Qt6 re https://github.com/MeVisLab/pythonqt/issues/13#issuecomment-1419390865?

iakov:
I'm not associated with the MeVisLab. Just a responsible user of the PythonQt.

usiems:
> Hi all! I’ve observed the recent development here in the PythonQt repo recently which is great to see 🙏🏻. Are you all associated with MeVisLab? Are you all prepping the testing infrastructure and modernizing the build infrastructure with the specific next step of updating to Qt6 re [#13 (comment)](https://github.com/MeVisLab/pythonqt/issues/13#issuecomment-1419390865)?

Me and @mrbean-bremen are members of the MeVisLab core developer team. And no, this is no direct preparation for Qt6 support - which is on our list, but not yet a pressing concern for us. It is just that my colleague got fed up with the current situation of neglect enough to do something about it. Many thanks for that! Obviously it inspired others to help out, too, which we are also thankful for.

PR #94, opened 2023-02-09
Use XSETREF for safety

iakov:
More details on [XSETREF macro for safer code](https://github.com/python/cpython/issues/99537)


mrbean-bremen (approved):
Looks good!

PR #95, opened 2023-02-09
Add debug builds into CI

iakov:
Fix UBSAN error for invalid access through pointer. This problem occurs only with Python2 because Py_DECREF is slightly different. Still, I have no opinion if UBSAN is right or wrong here, but this happens with the latest available XCode/clang too.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,8 +55,10 @@ jobs:
         echo ============================
         qmake -r PythonQt.pro CONFIG+=release CONFIG+=sanitizer CONFIG+=sanitize_undefined \
         PYTHON_VERSION=$(python3 --version | cut -d " " -f 2 | cut -d "." -f1,2) \
-        PYTHON_DIR=$(which python3 | xargs dirname | xargs dirname)    
-        make -j 2 && make check TESTARGS="-platform offscreen"
+        PYTHON_DIR=$(which python3 | xargs dirname | xargs dirname)
+        make -j 2
+        UBSAN_OPTIONS="halt_on_error=1" ASAN_OPTIONS="detect_stack_use_after_return=1:fast_unwind_on_malloc=0" \
+          make check TESTARGS="-platform offscreen"
       
     - name: Generate Wrappers
       run: |
@@ -86,6 +88,7 @@ jobs:
       matrix:
         container:
         - 'centos:7'
+        configuration: ['debug', 'release']
     runs-on: ubuntu-latest
     container: ${{ matrix.container }}
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -55,8 +55,10 @@ jobs:
         echo ============================
         qmake -r PythonQt.pro CONFIG+=release CONFIG+=sanitizer CONFIG+=sanitize_undefined \
         PYTHON_VERSION=$(python3 --version | cut -d " " -f 2 | cut -d "." -f1,2) \
-        PYTHON_DIR=$(which python3 | xargs dirname | xargs dirname)    
-        make -j 2 && make check TESTARGS="-platform offscreen"
+        PYTHON_DIR=$(which python3 | xargs dirname | xargs dirname)
+        make -j 2
+        UBSAN_OPTIONS="halt_on_error=1" ASAN_OPTIONS="detect_stack_use_after_return=1:fast_unwind_on_malloc=0" \
```
mrbean-bremen, line 60: I have no experience with UBSAN, will rely on your expertise here.

Unrelated note: I'm actually now thinking that these container based test on Ubuntu can be replaced tests based on `install-qt-action` as the macos and Windows tests - this makes it better configurable. Maybe only retain the CentOS one to be sure that it still works there...

src/PythonQtSignal.cpp
```diff
@@ -99,9 +99,10 @@ PythonQtSignalFunction_New(PythonQtSlotInfo *ml, PyObject *self, PyObject *modul
 /* Methods (the standard built-in methods, that is) */
 
 static void
-meth_dealloc(PythonQtSignalFunctionObject *m)
+meth_dealloc(PyObject *o)
```
mrbean-bremen, line 102: I don't see why this would change anything, but if it shuts up the checker...

mrbean-bremen (approved):
Looks good to me, thanks again!

.github/workflows/build.yml
iakov: I advise to save a previous and the latest Ubuntu LTS build in CI to test default build with them. Just to be sure that everything woks fine on typical developer machine.

src/PythonQtSignal.cpp
iakov: Probably, it is ([paragraph 7](https://en.cppreference.com/w/cpp/language/reinterpret_cast). And Python calls via pointer cast as `void (*)(PyObject*)`, thus it is an undefined behavior.

mrbean-bremen: That sounds right. I still don't think it will make a difference in practice, but you are right, the compiler is allowed to do anything in this case.

PR #103, opened 2023-05-24
Revert undefining some defines

mrbean-bremen:
- this has been introduced in #89 and breaks builds including *PythonQt.h*

The reverted code undefines `_POSIX_THREADS` (and a couple of other defines), which are queried in *PythonQt.h*.
If it is not set on a POSIX system, we get the error:
```Require native threads. See https://bugs.python.org/issue31370```

As far as I can see, none of these defines comes from *python.h*, at least not in current versions of Python.

### DIFF
--- a/src/PythonQtPythonInclude.h
+++ b/src/PythonQtPythonInclude.h
@@ -33,19 +33,6 @@
 #ifndef __PythonQtPythonInclude_h
 #define __PythonQtPythonInclude_h
 
-// Undefine macros that Python.h defines to avoid redefinition warning.
-#ifdef _POSIX_C_SOURCE
-#  undef _POSIX_C_SOURCE
-#endif
-
-#ifdef _POSIX_THREADS
-#  undef _POSIX_THREADS
-#endif
-
-#ifdef _XOPEN_SOURCE
-#  undef _XOPEN_SOURCE
-#endif
-
 // Undefine Qt keywords that conflict with Python headers
 #ifdef slots
 #undef slots


mrbean-bremen:
@iakov - please have a look.

iakov:
No objections. Next time I'll add the exact test to CI or an explanation into the code comment. This maybe the case with some mingw, however I have no chance to check until the weekend. 

But what Python version with what compiler produces the error in question? Maybe you can add the config to the CI (or just change/substitute the existing one)?

mrbean-bremen:
> But what Python version with what compiler produces the error in question? 

This happened in user code that included *PythonQt.h* right at the top of the file. It happened under gcc9, gcc11 and clang, with Python 9 and 11 (basically all tested POSIX systems). 

I just noticed that I wrote in the summary about *pthread.h* instead of *PythonQt.h* - just corrected that. *pythread.h* is where the error originates.

he-hesce:
@mrbean-bremen: Though it could be that even if _POSIX_THREADS caused issues, it could be that undefining _POSIX_C_SOURCE and _XOPEN_SOURCE was beneficial as now there is lots of redefinition warnings that pyconfig-64.h is redefining these two originally defined in from features.h. Undeffing them also seems to be okay as it doesn't break anything it seems. Code compiles and runs fine. It was only the undeffing of _POSIX_THREADS which caused compilation issues.

mrbean-bremen:
Ok, thanks - seem to have thrown out the baby with the bathwater... 
I'll check this later, and maybe just move the release tag after correcting it.

he-hesce:
@mrbean-bremen : Sounds okay.

PR #104, opened 2023-06-26
CI runners clean-up 

iakov:
1. macOS 10.15 was removed from GHA long ago.
https://github.blog/changelog/2022-07-20-github-actions-the-macos-10-15-actions-runner-image-is-being-deprecated-and-will-be-removed-by-8-30-22/
2. Python 2.7 is obsolete, thus removed for Windows and macOS, but still in CentOS7
3. RockyLinux9 is here for CI from now on

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -82,15 +82,21 @@ jobs:
         name: wrappers_ubuntu_22_04
         path: generated_cpp
 
-  centos:
+  oldschool:
     strategy:
       fail-fast: false
       matrix:
-        container:
-        - 'centos:7'
+        container_os: ['centos']
+        container_os_version: ['7']
+        container_os_python_package: ['python-debug']
         configuration: ['debug', 'release']
+        include:
+          - container_os: 'rockylinux'
+            container_os_version: '9'
+            container_os_python_package: 'python3-devel'
+            configuration: 'release'
     runs-on: ubuntu-latest
-    container: ${{ matrix.container }}
+    container: '${{ matrix.container_os }}:${{ matrix.container_os_version }}'
     steps:
     - name: Install Qt
       run: |
@@ -98,7 +104,7 @@ jobs:
         yum groupinstall "Development Tools" -y
         yum install -y \
         which \
-        python-debug \
+        ${{ matrix.container_os_python_package }} \
         qt5-qtbase-* \
         qt5-qttools* \
         qt5-qtsvg \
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -139,7 +139,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-       macos-version:  ['10.15']
+       macos-version:  ['11']
```
mrbean-bremen, line 142: I think the actual problem is the missing support for Python 2.7, also with the previous runner. I'm not sure how to test Python 2.7 on MacOS at all with that problem...

mrbean-bremen: Ok, and I just saw:
> Warning: The support for python 2.7 will be removed on June 19

for the Windows build, so the same problem here :(

iakov:
@mrbean-bremen  Seems like we have two different issues to address.
1. macOS 10.15 runner is outdated. And I propose to silently shift to the available one.
2. python 2.7 is unavailable for "low-effort" installation, but can be installed after some additional work.

What is your opinion, if there are any good reasons to keep CI builds with the Python 2.7 on modern OSes? Probably, the Centos7 build is enough to check a backward compatibility.

he-hesce:
As long as it continues to work with Python 2.7 on CentOS 7 as RHEL 7 is still in support window and in production use for a few more years.

he-hesce:
Though would be nice with test builds on Rocky 9 (RHEL 9) as that may be the next platform for those moving on up from CentOS 7 (RHEL 7) in the coming years.

mrbean-bremen:
> As long as it continues to work with Python 2.7 on CentOS 7 as RHEL 7 is still in support window and in production use for a few more years

As long as the docker container is available, this should not be a problem.

> Thought would be nice with test builds on Rocky 9 (RHEL 9) as that may be the next platform for those moving on up from CentOS 7 (RHEL 7) in the coming years once extended support expires. Rocky 9 would be Python 3.9.

Sure, this should be easy to setup if needed.

.github/workflows/build.yml
```diff
@@ -114,40 +120,37 @@ jobs:
...
         path: generated_cpp
 
   macOS:
     strategy:
       fail-fast: false
       matrix:
-       macos-version:  ['10.15']
-       python-version: ['2.7']
+       macos-version:  ['11']
+       python-version: ['3.6']
```
mrbean-bremen, line 150: Hopefully, 3.6 will be supported a bit longer by GH Actions, but eventually it will also go... Well we can deal with it when it becomes a problem.

iakov:
Ready for review (and squash-merge). The check fails because of `qt.io` download access problem, might be it can be just re-run once again.

.github/workflows/build.yml
```diff
@@ -82,23 +82,29 @@ jobs:
...
       fail-fast: false
       matrix:
-        container:
-        - 'centos:7'
+        container_os: ['centos']
+        container_os_version: ['7']
+        container_os_python_package: ['python-debug']
         configuration: ['debug', 'release']
+        include:
+          - container_os: 'rockylinux'
```
mrbean-bremen, line 94: Thanks - @he-hesce will be happy 😄

mrbean-bremen (approved):

he-hesce:
:-) I and everyone supporting production systems for 10-20 from original release will be happy.

PR #106, opened 2023-07-09
Add back undefining _POSIX_C_SOURCE and _XOPEN_SOURCE

mrbean-bremen:
- only undefining _POSIX_THREADS had be a problem
- see #103

### DIFF
--- a/src/PythonQtPythonInclude.h
+++ b/src/PythonQtPythonInclude.h
@@ -33,6 +33,15 @@
 #ifndef __PythonQtPythonInclude_h
 #define __PythonQtPythonInclude_h
 
+// Undefine macros that features.h defines to avoid redefinition warning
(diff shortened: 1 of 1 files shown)


he-hesce:
I think it's good!

he-hesce:
Gold star release!

mrbean-bremen:
Thanks, good to know!

he-hesce:
@mrbean-bremen: Although RHEL/CentOS 7 is our in-house platform and will slowly be moved to RHEL/Rocky 9 over the coming years. We skipped RHEL/Rocky 8 as it was just a possible upgrade path while waiting for 9. Ubuntu is just a backup platform in case RHEL/Rocky/AlmaLinux go (more than today) bonkers and we need a way out away from them - not that Ubuntu isn't bonkers themselves.

PR #113, opened 2023-10-02
Qt6 fixes (generator code)

jbowler:
These commits fix a number of cases where PythonQt is using features that were obsolete in QT 5.15 LTS.  This is just in the "generator" code - I can't get a build to get any further than generator/parser/parser.cpp yet so there will be other deprecated (and not removed) cases later in the build.

At this point parser.cpp fails to build because Qt6 did a rewrite of QtXml and removed all the "SAX" classes.  It looks like rewriting parser.cpp to match could be done blind and would work in earlier versions as the Qt6 recommendation is to use a class (QXmlStreamReader) which has existed since Qt 4.3.  Unfortunately its a rewrite and I'm not happy trying that; I don't know anything about the schema being used for one thing!

So I might try using Qt5Compat as a temporary fix to avoid the rewrite and see if further progress is possible without major or incompatible (with Qt5.15) changes.

### DIFF
--- a/build/common.prf
+++ b/build/common.prf
@@ -70,9 +70,8 @@ PYTHONQT_GENERATED_PATH = $$PWD/../generated_cpp
 }
 
 VERSION = 3.2.0
-greaterThan(QT_MAJOR_VERSION, 5) | greaterThan(QT_MINOR_VERSION, 9): CONFIG += c++11
 win32: CONFIG += skip_target_version_ext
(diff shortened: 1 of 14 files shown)


jbowler:
I'll do it in bits.  loadRelaxed was not introduced until 5.14 (the docs say this, but don't point out that load was only deprecated after that...)

PR #114, opened 2023-10-02
Qt6: initial changes to make compilation work

jbowler:
This is a minimalist set to pass the build tests (though I don't know about the 'nmake not found' failure).  It just fixes the multiple definitions of the compiler standard and the QStringList typedef problem.

### DIFF
--- a/src/src.pri
+++ b/src/src.pri
@@ -2,8 +2,6 @@ DEFINES +=  PYTHONQT_EXPORTS
 
 INCLUDEPATH += $$PWD
 
-CONFIG += c++11
-
 gcc:!no_warn:!clang:QMAKE_CXXFLAGS += -Wno-error=missing-field-initializers
 *-clang*:!no_warn:QMAKE_CXXFLAGS += -Wno-error=sometimes-uninitialized
 
--- a/build/common.prf
+++ b/build/common.prf
@@ -70,9 +70,8 @@ PYTHONQT_GENERATED_PATH = $$PWD/../generated_cpp
 }
 
 VERSION = 3.2.0
-greaterThan(QT_MAJOR_VERSION, 5) | greaterThan(QT_MINOR_VERSION, 9): CONFIG += c++11
 win32: CONFIG += skip_target_version_ext
-gcc|win32-clang-msvc:QMAKE_CXXFLAGS += -Wno-deprecated-declarations -Wuninitialized -Winit-self -ansi -pedantic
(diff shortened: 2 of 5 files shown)


jbowler:
Ok, so the stupid way github works is that I can't drop any more of the commits until you have taken this one, so I'll wait for that.

src/src.pri
```diff
@@ -2,8 +2,6 @@ DEFINES +=  PYTHONQT_EXPORTS
 
 INCLUDEPATH += $$PWD
 
-CONFIG += c++11
```
mrbean-bremen, line 5: Is this because we may actually need a newer version of the standard?

mrbean-bremen (approved):
Thanks!

jbowler:
On Mon, Oct 2, 2023 at 11:55 PM mrbean-bremen ***@***.***>
wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In src/src.pri
> <https://github.com/MeVisLab/pythonqt/pull/114#discussion_r1343590567>:
>
> > @@ -2,8 +2,6 @@ DEFINES +=  PYTHONQT_EXPORTS
>
>  INCLUDEPATH += $$PWD
>
> -CONFIG += c++11
>
> Is this because we may actually need a newer version of the standard?
>

Qt6 requires C++17, but "CONFIG += c++17" doesn't do that.  The explanation
is in the original commit log (see the pull request).   I'm not sure why I
can't see it on github, it looks like the two commits I had in the merge
request got combined into one and the detailed commit logs got lost (at
least I just did a git fetch and they're not in the current qt6 branch.)
Here's the paragraph from the Qt documentation:
c++17 C++17 support is enabled. *This option has no effect if the compiler
does not support C++17*, or can't select the C++ standard. By default,
support is enabled.c++11 is the same.  "This option has no effect if the
compiler does not support C++17"; so it requests it but doesn't require
it.  Instead Qt seems to rely on support in cmake (which I don't think you
use anywhere) to *require *particular standard support.

The hard-wired -ansi in QMAKE_CXXFLAGS turns off C++17 (and C++11) support
so I understand why it doesn't do anything with Qt6 and I think the fact
that it does with Qt5 is probably unexpected, or a feature/bug of the
implementation of qmake (qmake -qt=qt5).

You can find the original commit logs in the pull request:

https://github.com/MeVisLab/pythonqt/pull/114/commits

(Click on either of the commits on that page to see the original log).

mrbean-bremen:
Sorry, I missed your commit comment first - deleted the question after I noticed, but not fast enough
Note that I'm traveling, so I can check this only very sporadically.

PR #117, opened 2023-10-03
Qt6: qSort and qStableSort removed

jbowler:
The suggested replacements are std::stable_sort and std::sort which work (compile) in both Qt5 (5.15LTS) and Qt6.

### DIFF
--- a/generator/generator.cpp
+++ b/generator/generator.cpp
@@ -39,6 +39,7 @@
 **
 ****************************************************************************/
 
+#include <algorithm> // for std::stable_sort, std::sort
 #include "generator.h"
 #include "reporthandler.h"
 #include "fileout.h"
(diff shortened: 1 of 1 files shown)


jbowler:
I didn't add an include of <algorithm>, I assume the Qt headers are pulling it in but it may be that some of the older builds show a problem.  std::sort and std::stable_sort seem to have existed in C++ for ever (from cppreference.com).

mrbean-bremen:
> I didn't add an include of <algorithm>, I assume the Qt headers are pulling it in

That maybe, but I think it is still good practice to include them instead of relying on some incidendial include that may change with a newer Qt version

jbowler:
That's my preference too.  I'll push an addition to this pull (in a few hours)

mrbean-bremen (approved):

PR #120, opened 2023-10-05
Qt6: QTextStream compatibility

jbowler:
This provides compatibility with the removal of the global ::endl for use with QTextStream; Qt6 requires Qt::endl.

It also includes Qt6 specific changes necessary to deal with the replacement of QTextCodec by QStringConverter and the slight changes to the default setup of QTextStream (to UTF-8 with auto-detect of UTF-16).

### DIFF
--- a/generator/abstractmetabuilder.cpp
+++ b/generator/abstractmetabuilder.cpp
@@ -56,7 +56,9 @@
 #include <QtCore/QDebug>
 #include <QtCore/QFile>
 #include <QtCore/QFileInfo>
-#include <QtCore/QTextCodec>
+#if QT_VERSION < QT_VERSION_CHECK(6,0,0)
+#   include <QtCore/QTextCodec>
+#endif
 #include <QtCore/QTextStream>
 #include <QtCore/QVariant>
 
@@ -402,7 +404,10 @@ bool AbstractMetaBuilder::build()
         return false;
 
     QTextStream stream(&file);
-    stream.setCodec(QTextCodec::codecForName("UTF-8"));
+#   if QT_VERSION < QT_VERSION_CHECK(6,0,0)
+        stream.setCodec(QTextCodec::codecForName("UTF-8"));
+        /* Note required in Qt6: see the same call in asttoxml.cpp */
+#   endif
     QByteArray contents = stream.readAll().toUtf8();
     file.close();
 
--- a/generator/asttoxml.cpp
+++ b/generator/asttoxml.cpp
@@ -47,7 +47,9 @@
 
 #include <QXmlStreamWriter>
 #include <QTextStream>
-#include <QTextCodec>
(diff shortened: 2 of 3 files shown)


generator/typesystem.h
```diff
@@ -58,6 +61,27 @@
...
 #endif
 
+/* Global endl (::endl) is used extensively in the generator .cpp files.  This
+ * was supported by Qt until Qt6.  In Qt5.14 Qt::endl was added in anticipation
+ * of the Qt6 change and the use of global endl could be avoided (it does not
+ * seem to have been explicitly deprecated).  This gives backward compatibility
+ * for global endl in Qt6 (not Qt5, where global endl was still available).
+ *
+ * Note that 'constexpr' is available in Qt6 because Qt6 requires C++17;
+ * consexpr was introduced in C++11.  Likewise for decltype.  Qt::endl is a
```
mrbean-bremen, line 71: Nitpicking - typo constexpr

jbowler: Typo?  It seems correct to me; it's spelled the same in the code and the code compiles!

jbowler: Oh... sorry; the second one...

mrbean-bremen (approved):

generator/abstractmetabuilder.cpp
```diff
@@ -402,7 +404,10 @@ bool AbstractMetaBuilder::build()
         return false;
 
     QTextStream stream(&file);
-    stream.setCodec(QTextCodec::codecForName("UTF-8"));
+#   if QT_VERSION < QT_VERSION_CHECK(6,0,0)
+        stream.setCodec(QTextCodec::codecForName("UTF-8"));
+        /* Note required in Qt6: see the same call in asttoxml.cpp */
+#   endif
     QByteArray contents = stream.readAll().toUtf8();
```
usiems, line 411: Why is a QTextStream used here anyway, if the read content is converted back to UTF-8 immediately?
It seems to me the content of the file could be read without the intermediate decoding/encoding step.

generator/asttoxml.cpp
```diff
@@ -57,7 +59,18 @@ void astToXML(QString name) {
...
+    /* NOTE, for Qt6:
+     *
+     *  stream.setEncoding(QStringConverter::Utf8)
+     *
+     * is the default but will be overridden if the UTF-16 BOM is seen.  This
+     * is almost certainly the correct behavior because the BOM isn't valid in
+     * a text stream otherwise.
+     */
+#endif
     QByteArray contents = stream.readAll().toUtf8();
```
usiems, line 74: Same here.

generator/abstractmetabuilder.cpp
jbowler: Maybe to remove the BOM which would probably mess up the resultant byte array?  It also handles UTF-16 correctly in Qt6.

PR #121, opened 2023-10-05
Eliminate warnings about anonymous structs on GCC

jbowler:
Corresponding warnings are already turned off on clang in generator/generator.pri.  There's no G++ command-line option to turn them off without explicitly disabling other warnings (-Wpedantic).

### DIFF
--- a/generator/parser/codemodel.h
+++ b/generator/parser/codemodel.h
@@ -517,16 +517,16 @@ private:
   {
     struct
     {
-      uint _M_isConstant: 1;
-      uint _M_isConstexpr: 1;
-      uint _M_isVolatile: 1;
-      uint _M_isStatic: 1;
-      uint _M_isAuto: 1;
-      uint _M_isFriend: 1;
-      uint _M_isRegister: 1;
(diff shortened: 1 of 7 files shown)


generator/parser/codemodel.h
```diff
@@ -512,6 +517,9 @@ class _MemberModelItem: public _CodeModelItem
       uint _M_isExtern: 1;
       uint _M_isMutable: 1;
     };
+#if defined(__GNUC__)
+#pragma GCC diagnostic pop
+#endif
     uint _M_flags;
   };
```
mrbean-bremen, line 524: Wouldn't it be better to just use an allowed construct, e.g. something like
```cpp
union
{
    struct bitflags_type
    {
        uint _M_isConstant: 1;
        uint _M_isExtern: 1;
        uint _M_isMutable: 1;
    } bitflags;
    uint _M_flags;
};
```

jbowler: Changed to give the structs names rather than disabling the warnings.

mrbean-bremen (approved):
Thanks!

PR #123, opened 2023-10-05
Qt6: generator/typesystem.cpp: XML workround

jbowler:
This works around the need to rewrite the XML parser in generator/typesystem.cpp by invoking the Qt5 compatibility library; this is the only thing in the generator which requires it but the alternative is a complete rewrite of the XML reading code.

### DIFF
--- a/generator/typesystem.cpp
+++ b/generator/typesystem.cpp
@@ -51,6 +51,22 @@
 #include <QtXml>
 #include <qcompilerdetection.h> // Q_FALLTHROUGH
 
+/* This file needs to be rewritten as documented here:
+ *
+ * See: https://doc.qt.io/qt-6/xml-changes-qt6.html
+ *
+ * The rewrite may be backward compatible to Qt4.3 APIs because the base
+ * facilites (QXmlStreamReader) used to relace the 'SAX' parser were apparently
+ * available then.  Use of Xml5Compat is a work round until such a rewrite has
+ * been done.
+ */
+#if QT_VERSION >= QT_VERSION_CHECK(6,0,0)
+#   if defined(__GNUC__)
+#       pragma GCC warning "Qt6: implement Qt6 compatible XML reading"
+#   endif
+#   include <QtCore5Compat/QXmlDefaultHandler>
+#endif
+
 QString strings_Object = QLatin1String("Object");
 QString strings_String = QLatin1String("String");
 QString strings_Thread = QLatin1String("Thread");
--- a/generator/generator.pri
+++ b/generator/generator.pri
@@ -70,6 +70,11 @@ SOURCES += \
    
 QT += core xml
 
+greaterThan(QT_MAJOR_VERSION, 5) {
+    QT += core5compat
+    message("WARNING: Qt module core5compat for XML handling in typesystem.cpp")
+}
+
 win32-msvc.net {
         QMAKE_CXXFLAGS += /Zm500
         QMAKE_CXXFLAGS -= -Zm200


jbowler:
with this change generator/ can build against Qt6.  Next step is to get the generated files.

generator/typesystem.cpp
```diff
@@ -51,6 +51,22 @@
...
+ * facilites (QXmlStreamReader) used to relace the 'SAX' parser were apparently
+ * available then.  Use of Xml5Compat is a work round until such a rewrite has
+ * been done.
+ */
+#if QT_VERSION >= QT_VERSION_CHECK(6,0,0)
+#   if defined(__GNUC__)
+#       pragma GCC warning "Qt6: implement Qt6 compatible XML reading"
+#   endif
+#   include <QtCore5Compat/QXmlDefaultHandler>
+#endif
```
mrbean-bremen, line 68: As mentioned by @usiems , the plan is to replace the generator/parser for Qt6 with a shiboken-derived one, so this will only be a stopgap measure until then.

mrbean-bremen (approved):
Are you also planning to add CI tests for Qt6? Would be helpful...

jbowler:
> Are you also planning to add CI tests for Qt6? Would be helpful...

Unless it's possible to obtain working generated cpp files the src build doesn't complete.  The generator needs changes to recognize constexpr at least; I have those changes.  However the generator/typesystem_*.xml needs fixing for both 5.15LTS and 6.5LTS and debugging what is failing is hard because of the error reporting (errors are reported in the output of pythonqt_generator but there is no clue as to which Qt header is causing them, then in the src/ build the misgenerated files produce obvious errors.)

mrbean-bremen:
True. I'm a bit unsure how to handle compatibility in this case. The idea for Qt6 is to use a separate generator based on shiboken, including separate typesystem files, but I don't know to handle the changes in Qt 5.15 in a backwards compatible way. 

I don't think there is a possibility to add version-dependent entries in the typesystem files, so maybe we need to have separate typesystem files for separate versions, at least for pre- and post 5.15 versions. The generator would have be to adapted to use the correct typesystem files depending on the Qt version. Adapting the generator to also correctly parse Qt6 would probably be too much - @florianlink mentioned that it should be easier to use a shiboken-based generator instead of manually having to adapt the current one, and he knows the code best.

jbowler:
> I don't think there is a possibility to add version-dependent entries in the typesystem files.

Some previous contributions seemed to be trying to do that:

https://github.com/MeVisLab/pythonqt/commit/58ccd5393922d88fd320628aa2db787df1f1233b

But XML does not define  as anything other than a comment so what that commit does is raise the question of which one of the lines is being used.  It's what I'm working on at present; the error/warning reporting from pythonqt_generator is not good enough to fix the typesystem files in finite time.

So maybe we need to have separate typesystem files for separate versions, at least for pre- and post 5.15 versions. 

I think so.  I started changing them in-place then stopped (because of the error message problem).  I've got Qt5.15 and Qt6 changes in new subdirectories generator/typesystem_515 and generator/typesystem_6.  I'm not sure yet what I need to change to get a particular build to use those if they exist and/or whether I can build a search path.

>The generator would have be to adapted to use the correct typesystem files depending on the Qt version. Adapting the >generator to also correctly parse Qt6 would probably be too much

So far as the parsing is concerned it just looks like the missing handling for 'constexpr' on functions.

It just depends on what the timescale is for the shiboken version; there's not a lot of point during 5.15LTS at present as 5.11 sort-of-works.

mrbean-bremen:
> Some previous contributions seemed to be trying to do that:

I think that are really just comments that are not used for parsing.

>  I'm not sure yet what I need to change to get a particular build to use those if they exist and/or whether I can build a search path.

I haven't looked at this either, but it shouldn't be too complicated, given that the xml files have to be read in from a specific path at some time in the generator code. I may have a closer look later.

> So far as the parsing is concerned it just looks like the missing handling for 'constexpr' on functions.

If this is indeed the case, we may just forgo the shiboken path and use the current parser - that would be be way easier. I had a closer look at shiboken, and at a first glance it did seem quite a bit of work to adapt it, at least to me.

jbowler:
> I haven't looked at this either, but it shouldn't be too complicated, given that the xml files have to be read in from a specific path at some time in the generator code. I may have a closer look later.

They aren't read from files by the generator; they are resources.   See my next pull request; it took me forever (well, a few days) to work out why changing the .xml files didn't change the output of the generator.  The changes only take effect after a "make" within generator/

I was intending to supply a fix.  I don't think I can do a search path within qmake but I'm pretty sure I can find the subdirectories I specified.  My proposal if I can do it is to move all the required files into subdirectories (all the .xml files, build-all.txt and the master cpp file) and choose the right one within the build.

mrbean-bremen:
Ok - I was clearly too optimistic. Thanks for your time and effort!

usiems:
> I don't think there is a possibility to add version-dependent entries in the typesystem files, so maybe we need to have separate typesystem files for separate versions, at least for pre- and post 5.15 versions. The generator would have be to adapted to use the correct typesystem files depending on the Qt version.

It also might be possible to introduce reading version specifiers from the XML tags, something like `qtVersionGreaterEqualThan="5.15"` or `qtVersionLessThan="6"` (if we want to be really verbose), and every tag where the version doesn't fit is simply dropped. If this is a workable strategy depends on how many tags would need to be version specific. I imagine that there is still great overlap between Qt 6 and Qt 5, and classes and methods that are not in the header files (anymore) are ignored anyway. At least this would spare us to copy the typesystem files - though starting from scratch would also have advantages.

PR #124, opened 2023-10-09
pythonqt_generator modified argument handling, make generate

jbowler:
This adds 'make generate' to generator/generator.pro using modified argument handling so that the Qt installed header directory can be used to build a new generated_cpp.  This is created in generator/ to avoid overwriting an existing user generated_cpp at the top level.

It also adds setContext calls to ReportHandler to improve the warning messages and calls fflush(stdout) at appropriate moments to fix the confusing mis-ordering of the pythonqt_generator output when it is piped to a file.

### DIFF
--- a/generator/main.cpp
+++ b/generator/main.cpp
@@ -246,13 +246,8 @@ int main(int argc, char *argv[])
 
         /* Warn if standard directories are not found: */
         if (!requiredModules.isEmpty()) {
-            QString sep("");
-            QString errMsg("WARNING: missing core Qt modules:");
-
-            for (const QString &mod : requiredModules) {
-                errMsg += sep + " Qt" + mod;
-                sep = ",";
-            }
+            const QString errMsg("WARNING: missing core Qt modules: Qt" +
+                    requiredModules.join(",Qt"));
 
             if (args.contains("core-error"))
                 qFatal(errMsg.toLocal8Bit().constData());
--- a/generator/main.cpp
+++ b/generator/main.cpp
@@ -187,7 +187,7 @@ int main(int argc, char *argv[])
             qWarning(
 "QTDIR environment variable not set. Assuming standard binary install using\n"
 "frameworks.");
-            foreach (const QString &mod, coreModules)
+            for (const QString &mod : coreModules)
                 includes << ("/Library/Frameworks/Qt" + mod + ".framework/Headers");
             includes << "/Library/Frameworks"; // this seems wrong
 #else
@@ -220,11 +220,11 @@ int main(int argc, char *argv[])
         QStringList dirList;
         QStringList requiredModules(coreModules);
 
-        foreach (const QString &dir, qtHeaders.split(QDir::listSeparator()))
+        for (const QString &dir : qtHeaders.split(QDir::listSeparator()))
             if (QDir(dir).exists())  {
                 QStringList remaining(requiredModules);
 
-                foreach (const QString &mod, requiredModules) {
+                for (const QString &mod : requiredModules) {
                     const QString modpath(dir + "/Qt" + mod);
                     if (QDir(modpath).exists()) {
                         includes << modpath;
@@ -249,7 +249,7 @@ int main(int argc, char *argv[])
             QString sep("");
             QString errMsg("WARNING: missing core Qt modules:");
 
-            foreach (const QString &mod, requiredModules) {
+            for (const QString &mod : requiredModules) {
                 errMsg += sep + " Qt" + mod;
                 sep = ",";
             }
@@ -269,7 +269,7 @@ int main(int argc, char *argv[])
 
     std::cout << "-------------------------------------------------------------" << std::endl;
     std::cout << "Scanning Qt headers from (in this order):" << std::endl;
-    foreach (const QString &dir, includes) {
+    for (const QString &dir : includes) {
         std::cout << "  " << dir.toUtf8().constData();
         if (!QDir(dir).exists())
             std::cout << " [DIRECTORY DOES NOT EXIST]";
--- a/generator/main.cpp
+++ b/generator/main.cpp
@@ -149,21 +149,15 @@ int main(int argc, char *argv[])
     if (!minimal)
         includes << QString(".");
 
-#if defined(Q_OS_WIN32)
-    const char *path_splitter = ";";
-#else
-    const char *path_splitter = ":";
-#endif
-
     // Environment INCLUDE
     const QString includePath(getenv("INCLUDE"));
     if (!includePath.isEmpty())
-        includes += includePath.split(path_splitter);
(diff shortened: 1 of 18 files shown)


mrbean-bremen:
Thanks - I will check this tomorrow (too sleepy now...)!

generator/main.cpp
```diff
@@ -132,36 +137,205 @@ int main(int argc, char *argv[])
...
+        if (!minimal)
+            includes += dirList;
+
+        /* Warn if standard directories are not found: */
+        if (!requiredModules.isEmpty()) {
+            QString sep("");
+            QString errMsg("WARNING: missing core Qt modules:");
+
+            foreach (const QString &mod, requiredModules) {
+                errMsg += sep + " Qt" + mod;
```
mrbean-bremen, line 259: Probably easier to add the "Qt" right at the beginning and use `join(", ")` here...

generator/main.cpp
```diff
@@ -132,36 +137,205 @@ int main(int argc, char *argv[])
...
+                qFatal(errMsg.toLocal8Bit().constData());
+            else
+                qWarning(errMsg.toLocal8Bit().constData());
+        }
+    } else if (includes.isEmpty()) {
+        /* This is never true unless --minimal is given because "." is always
+         * added to includes first.
+         */
+        qWarning("No directories to scan: use --include-paths or --qt-headers");
+        displayHelp(&*gs);
```
mrbean-bremen, line 273: This looks weird (I know, it is used elsewhere), I would use `displayHelp(gs.get())` for clarity instead. Matter of taste, I guess...
Also, while at it - probably better to call the function something like `displayHelpAndExit`.

generator/main.cpp
```diff
@@ -132,36 +137,205 @@ int main(int argc, char *argv[])
...
+        if (!QDir(dir).exists())
+            std::cout << " [DIRECTORY DOES NOT EXIST]";
+        std::cout << std::endl;
+    }
+    std::cout << "-------------------------------------------------------------" << std::endl;
+
     printf("Please wait while source files are being generated...\n");
 
     printf("Parsing typesystem file [%s]\n", qPrintable(typesystemFileName));
+    fflush(stdout);
```
mrbean-bremen, line 289: Note that `std::endl` also does a flush, though I guess `printf` was used here throughout the code, and you don't want to replace it all...

mrbean-bremen (commented):
Looks quite good to me, thanks! @usiems - could you also have a look?

usiems:
Looks good to me so far.

generator/main.cpp
```diff
@@ -132,36 +137,205 @@ int main(int argc, char *argv[])
...
+    const bool minimal(args.contains("minimal"));
+    QStringList includes;
+    if (!minimal)
+        includes << QString(".");
+
+#if defined(Q_OS_WIN32)
+    const char *path_splitter = ";";
+#else
+    const char *path_splitter = ":";
+#endif
```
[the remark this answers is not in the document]
jbowler: Ok; I was trying to minimise the changes (the lines above are in the original).

generator/main.cpp
```diff
@@ -132,36 +137,205 @@ int main(int argc, char *argv[])
...
+    if (!commandLineIncludes.isEmpty())
+        includes += commandLineIncludes.split(path_splitter);
+
+    /* Allow coreModules to be specified on the command line (e.g. to add or
+     * remove modules for an application specific build).
+     */
+    QString moduleList(args.value("core-modules"));
+    if (moduleList.isEmpty())
+        moduleList = coreModuleListDefault;
+    const QStringList coreModules(moduleList.split(","));
```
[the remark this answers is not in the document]
jbowler: It's more for testing/error recovery (if this patch gets accepted).  The generator doesn't use the list to limit what it includes, it just silently ignores modules that aren't found.  The list only controls what include directories get searched (by the #includes in qtscript_masterinclude.h).  --minimal serves the same purpose by removing the top-level-directory from the search list.

generator/main.cpp
```diff
@@ -132,36 +137,205 @@ int main(int argc, char *argv[])
...
+         * NOTE: use of a list here is intended for the project build workflows
+         * which apparently have to assemble the headers from difference
+         * installation directories.  This could be avoided by using
+         * --include-paths instead!
+         *
+         */
+        QStringList dirList;
+        QStringList requiredModules(coreModules);
+
+        foreach (const QString &dir, qtHeaders.split(path_splitter))
```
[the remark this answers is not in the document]
jbowler: My preference too.  I didn't know if -ansi support was still required.

Shouldn't take long.

generator/main.cpp
jbowler: I always follow the conventions of the existing code when submitting patches.  Perhaps it might be better to do code cleanup as a separate task?

mrbean-bremen: Agreed - you are right, of course.

generator/main.cpp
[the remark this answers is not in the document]
mrbean-bremen: Yes, range loops are also used in the rest of the code.

generator/main.cpp
mrbean-bremen: > the baroque nature of the rest of the code

I like that 😁

generator/main.cpp
```diff
@@ -246,13 +246,8 @@ int main(int argc, char *argv[])
...
         if (!requiredModules.isEmpty()) {
-            QString sep("");
-            QString errMsg("WARNING: missing core Qt modules:");
-
-            for (const QString &mod : requiredModules) {
-                errMsg += sep + " Qt" + mod;
-                sep = ",";
-            }
+            const QString errMsg("WARNING: missing core Qt modules: Qt" +
+                    requiredModules.join(",Qt"));
```
mrbean-bremen, line 250: Nitpicking: If you want it as before (with the space) you need `join(", Qt"))`.

usiems (approved):

jbowler:
This is getting too complicated and there is too much change of a merge conflict with other forks.

mrbean-bremen:
> This is getting too complicated and there is too much change of a merge conflict with other forks.

Hm, I was actually going to merge this... this last nitpicking was not really important.
Do you want to split it, or just drop it?

jbowler:
Just drop it; I've convinced myself that the core-modules stuff is irrelevant because it, in fact, does nothing.  I doubt any system really installs Qt headers with a repeated module directory, e.g. $QTDIR/QtCore/QtCore/qcoreapplication

I can run pythonqt_generator by hand and get the same results.  The real changes I need are in the error reporting; I only submitted this PR because I did it first so I'm going to have to spend some time de-merging the changes in my own branches/stashes.

mrbean-bremen:
Ok - in case we still decide that the functionality is needed, we can recreate it from this PR... thanks!

PR #125, opened 2023-10-10
constexpr and decltype parsing

jbowler:
Parses constexpr (in the manner of const) and decltype (in the manner of __typeof).  Adds flags for them and also for other already parsed items (mutual, auto).


generator/parser/type_compiler.cpp
```diff
@@ -127,6 +127,7 @@ void TypeCompiler::visitName(NameAST *node)
   _M_type = name_cc.qualifiedName();
 }
 
+#if 0 //UNUSED
 QStringList TypeCompiler::cvString() const
 {
   QStringList lst;
```
mrbean-bremen, line 133: You may just remove this.

generator/parser/type_compiler.h
```diff
@@ -61,9 +61,11 @@ class TypeCompiler: protected DefaultVisitor
   inline QList<int> cv() const { return _M_cv; }
 
   bool isConstant() const;
+  bool isConstexpr() const;
   bool isVolatile() const;
+  bool isMutable() const;
 
-  QStringList cvString() const;
+  //QStringList cvString() const; //UNUSED
```
mrbean-bremen, line 68: Just remove it - it was never used.

generator/parser/codemodel.cpp
```diff
@@ -618,6 +626,10 @@ bool _FunctionModelItem::isSimilar(FunctionModelItem other) const
   if (isConstant() != other->isConstant())
     return false;
 
+  // ### this may be incorrect (note that isVolatile is not checked!)
+  if (isConstexpr() != other->isConstexpr())
+    return false;
+
```
mrbean-bremen, line 632: I think this is not needed, as `constexpr` functions cannot be overloaded.

mrbean-bremen (commented):
Thank! Seems to be ok, as far as I can see, though I may miss something of course. The proof of the pudding... @usiems - please have a look at this, too.

usiems (commented):
Note that, if new keywords are, for our purposes, equivalent to recognized ones, one might probably as well add a `#define` to the pp-qt-configuration file. The embedded preprocessor should not care.

Not directly related to this commit, but because support for `auto` was prepared in the code:
If `auto` is used for return types in the Qt headers, the generator has a problem, because the parser will not automatically derive the returned type, which in turn means that the correct code can't be generated.

This should be seldom be the case in Qt though, since this would mean that a method implementation is in the header file, which would somewhat defeat the promised stable binary interface. The only valid use case would be for templated container types, which PythonQt doesn't wrap anyway. Still, something we should be aware of.

generator/parser/codemodel.h
```diff
@@ -167,18 +179,25 @@ struct TypeInfo
 private:
   struct TypeInfo_flags {
 	uint m_constant: 1;
+	uint m_constexpr: 1;
 	uint m_volatile: 1;
+        uint m_mutable: 1;
+        uint m_friend: 1;
+        uint m_auto: 1;
```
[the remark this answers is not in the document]
jbowler: Tabs are messed up in diffs, but it looks like you are using 'expandtabs' and that is not the default in vim.  I'll push a fix.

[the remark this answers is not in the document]
jbowler: Removed friend and auto from codemodel.h

generator/parser/type_compiler.cpp
jbowler: Done

generator/parser/type_compiler.h
jbowler: Done

jbowler:
> If `auto` is used for return types in the Qt headers, the generator has a problem, because the parser will not automatically derive the returned type, which in turn means that the correct code can't be generated.

It is used in that way in Qt6.5.3, which is why I added it.  It's like decltype() or __typeof().  Typeof is ignored which means the type will be wrong.  'auto' might work if preserved.  I simply don't use C++ class auto so I'm not familiar with the restrictions.  decltype is likely to occur in templates (function templates maybe) which are not currently handled.

No problem until Qt6 then the function will just have to be disabled in the relevant typesystem file.  I added auto mainly so I could add big warning message when it is encountered.

generator/parser/codemodel.cpp
jbowler: Removed

usiems (approved):

mrbean-bremen (approved):

PR #127, opened 2023-10-13
Qt6: fixes for removal of Qt5 types

jbowler:
Also add the alternate types that are available in Qt5 to replace the types deprecated in Qt5.14 (allows Qt5.15LTS users to make changes for compatibility with Qt6.)


mrbean-bremen (approved):
Thanks!

PR #128, opened 2023-10-13
Fix problems in Qt5.11 generator

jbowler:
The fix for Qt5.12 breaks pythonqt_generator for 5.11.  Patch includes better checking for precompiled headers and errors out early rather than failing when linking the src/ with mysterious missing symbols.

### DIFF
--- a/build/common.prf
+++ b/build/common.prf
@@ -25,48 +25,34 @@ PYTHONQT_GENERATED_PATH = $$PWD/../generated_cpp
   PYTHONQT_GENERATED_PATH = $$PWD/../generated_cpp_$${QT_MAJOR_VERSION}$${QT_MINOR_VERSION}
 
   !exists($$PYTHONQT_GENERATED_PATH) {
-    contains( QT_MAJOR_VERSION, 5 ) {
-      contains( QT_MINOR_VERSION, 10 ) { 
-        PYTHONQT_GENERATED_PATH = $$PWD/../generated_cpp_56
-      }
-      else:contains( QT_MINOR_VERSION, 11 ) { 
-        PYTHONQT_GENERATED_PATH = $$PWD/../generated_cpp_511
-      }
-      else:contains( QT_MINOR_VERSION, 12 ) {
-        PYTHONQT_GENERATED_PATH = $$PWD/../generated_cpp_511
-      }
-      else:contains( QT_MINOR_VERSION, 1 ) {
+    # For Qt5 we know that the older generated wrappers work with the later
+    # versions, even (apparently) Qt5.15, so:
+    equals(QT_MAJOR_VERSION, 5) {
+      # Qt5: have 5.0, 5.3, 5.4, 5.6 and 5.11 at present:
+      lessThan(QT_MINOR_VERSION, 3) { # 5.1, 5.2
         PYTHONQT_GENERATED_PATH = $$PWD/../generated_cpp_50
       }
-      else:contains( QT_MINOR_VERSION, 2 ) {
-        PYTHONQT_GENERATED_PATH = $$PWD/../generated_cpp_50
-      }
-  	  else:contains( QT_MINOR_VERSION, 3 ) { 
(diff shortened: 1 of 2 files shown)


jbowler:
Florian Link's last check-in to generator/qtscript_masterinclude.h breaks pythonqt_generator on earlier versions than Qt5.12 (it introduces a duplicate definition because the #define results in declarations of the same functions in two places.)

Tested against Qt5.11.11 using header files generatored from Qt5.11.3, Qt5.12.12 and Qt5.13.2 (Qt5.14.2 generation fails for the same reason as Qt5.11.11 does; the Qt6 coompat changes.)

Also makes the build stop if the precompiled headers aren't found rather than wading through the complication then failing at the link step.

build/common.prf
```diff
@@ -25,50 +25,36 @@ PYTHONQT_GENERATED_PATH = $$PWD/../generated_cpp
...
-        PYTHONQT_GENERATED_PATH = $$PWD/../generated_cpp_56
+      else { # >5.11
+        # LATEST Qt5 generated files:
+        PYTHONQT_GENERATED_PATH = $$PWD/../generated_cpp_511
       }
     }
+
+    !exists($$PYTHONQT_GENERATED_PATH) {
+      error("No precompiled headers exist for Qt$${QT_VERSION}")
+    }
```
mrbean-bremen, line 49: Shouldn't that be "generated sources" instead of "precompiled headers"? Same below.

generator/qtscript_masterinclude.h
```diff
@@ -48,11 +47,29 @@
...
+/* This must only be included after 'QT_NO_' definitions have been defined. */
+#include <QtCore/qglobal.h>
+
+/* NOTE: Qt5.12 and later (including Qt6) uses template functions for the
+ * static implementations of QTimer::singleShot() (the function, not the
+ * property).  The generator does not handle template functions so this work
+ * round is necessary.  Without this the QTimer::singleShot functions do not
+ * appear in the PythonQt interface.
+ *
+ * Unfortunately the work round breaks precompilation in Qt5.11 because it
```
mrbean-bremen, line 59: Work round - > workaround

mrbean-bremen (commented):
We really need some tests to avoid this kind of regression...

build/common.prf
jbowler: Yes.

generator/qtscript_masterinclude.h
jbowler: Both are used.

Pushed changes.

jbowler:
> We really need some tests to avoid this kind of regression...

generated_cpp_512 was never added; do you know why?  Qt5.15.11 compiles and passes the tests with these headers so I could do a pull request with that and even with generated_cpp_513 but there's no point if there is some known problem with those generated files.

Adding tests for the generator requires the official Qt header files for the release.  Perfectly possible with a pre-built dedicated test machine but you seem to be building the system (in a docker or similar) each time so that's pretty tricky given that Qt changes the downloads regularly.

mrbean-bremen:
> generated_cpp_512 was never added; do you know why?

I guess it wasn't needed by anybody? Maybe @usiems knows the answer.

Generally, I think it would be nice if the sources could just be generated in a build step for the currently used version by the user, but at the moment it probably would make sense to add the generated sources at least for 5.12 and 5.15 (as soon as that works reliably).

mrbean-bremen (approved):
As a side note: the goal is to use cmake instead of qmake at least for Qt6, but this is something that can be added later.

usiems:
> > generated_cpp_512 was never added; do you know why?
> 
> I guess it wasn't needed by anybody? Maybe @usiems knows the answer.

The generator didn't work for the 5.12 sources, and I didn't invest the time to find and fix the problem, assuming that the differences between 5.11 and 5.12 wouldn't be that great. The same happened when we switched [MeVisLab](https://www.mevislab.de) to Qt 5.15, where the old 5.11 wrappers still are "good enough" for our purposes. Too much other stuff with higher priority.

PR #129, opened 2023-10-14
Retab

jbowler:
Expand all the tabs in non-binary source files.

### DIFF
--- a/generator/abstractmetalang.cpp
+++ b/generator/abstractmetalang.cpp
@@ -1203,7 +1203,7 @@ bool AbstractMetaClass::hasVirtualDestructor() const
 static bool functions_contains(const AbstractMetaFunctionList &l, const AbstractMetaFunction *func)
 {
     foreach (const AbstractMetaFunction *f, l) {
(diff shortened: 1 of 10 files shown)


jbowler:
The change for build/common.prf is in pull request #128

generator/abstractmetalang.cpp
```diff
@@ -1203,7 +1203,7 @@ bool AbstractMetaClass::hasVirtualDestructor() const
 static bool functions_contains(const AbstractMetaFunctionList &l, const AbstractMetaFunction *func)
 {
     foreach (const AbstractMetaFunction *f, l) {
-		if ((f->compareTo(func) & AbstractMetaFunction::PrettySimilar) == AbstractMetaFunction::PrettySimilar)
+                if ((f->compareTo(func) & AbstractMetaFunction::PrettySimilar) == AbstractMetaFunction::PrettySimilar)
```
mrbean-bremen, line 1206: Looks like here a tab should be 4 spaces - not that there is any consistent formatting here...

mrbean-bremen (approved):
Thanks. There remain inconsistencies in the formatting, but these have been there in the first place (as can be seen in the [original code](https://github.com/qt-labs/qtscriptgenerator)), so I'll merge this is as.

PR #131, opened 2023-10-15
generator: improve error handling

jbowler:
This ensures that printf() in the generator does not get file buffered (and therefore delayed relative to stderr) when piping output to a file and provides ReportHandler "context" for ReportHandler warnings.

### DIFF
--- a/generator/main.cpp
+++ b/generator/main.cpp
@@ -39,6 +39,8 @@
 **
 ****************************************************************************/
 
+#include <cstdio>
+
 #include "main.h"
 #include "asttoxml.h"
 #include "reporthandler.h"
@@ -53,6 +55,8 @@ void displayHelp(GeneratorSet *generatorSet);
 #include <QDebug>
 int main(int argc, char *argv[])
 {
+    ReportHandler::setContext("Arguments");
+
     QScopedPointer<GeneratorSet> gs(GeneratorSet::getInstance());
 
     QString default_file = ":/trolltech/generator/qtscript_masterinclude.h";
@@ -135,11 +139,13 @@ int main(int argc, char *argv[])
     printf("Please wait while source files are being generated...\n");
 
     printf("Parsing typesystem file [%s]\n", qPrintable(typesystemFileName));
+    ReportHandler::setContext("Typesystem");
     if (!TypeDatabase::instance()->parseFile(typesystemFileName))
         qFatal("Cannot parse file: '%s'", qPrintable(typesystemFileName));
 
     printf("PreProcessing - Generate [%s] using [%s] and include-paths [%s]\n",
       qPrintable(pp_file), qPrintable(fileName), qPrintable(args.value("include-paths")));
+    ReportHandler::setContext("Preprocess");
     if (!Preprocess::preprocess(fileName, pp_file, args.value("include-paths"))) {
         fprintf(stderr, "Preprocessor failed on file: '%s'\n", qPrintable(fileName));
         return 1;
@@ -148,16 +154,19 @@ int main(int argc, char *argv[])
     if (args.contains("ast-to-xml")) {
       printf("Running ast-to-xml on file [%s] using pp_file [%s] and include-paths [%s]\n",
         qPrintable(fileName), qPrintable(pp_file), qPrintable(args.value("include-paths")));
+      ReportHandler::setContext(QString("AST-to-XML"));
       astToXML(pp_file);
       return 0;
     }
 
     printf("Building model using [%s]\n", qPrintable(pp_file));
+    ReportHandler::setContext("Build");
     gs->buildModel(pp_file);
     if (args.contains("dump-object-tree")) {
         gs->dumpObjectTree();
         return 0;
     }
+    ReportHandler::setContext("Generate");
     printf("%s\n", qPrintable(gs->generate()));
 
     printf("Done, %d warnings (%d known issues)\n", ReportHandler::warningCount(),


mrbean-bremen (approved):
LGTM

mrbean-bremen:
Not sure what happens with the Windows build - it run into an STATUS_STACK_BUFFER_OVERRUN error...

jbowler:
> Not sure what happens with the Windows build - it run into an STATUS_STACK_BUFFER_OVERRUN error...

That sounds like infinite recursion.  I'll pushed a no-op change to force the builds to run again.

jbowler:
Ok, this time it succeeded.  Maybe setvbuf(stdout,nullptr,_IOLBF,0) is calling something recursively.  I'll try commenting it out.  Passing "nullptr" may be a problem; it's a C API not a C++ one.  "0" is safer.

jbowler:
It works without setvbuf so apparently there is some setvbuf bug with MSVC, or maybe the header files manage to do something weird.  I've re-pushed it with nullptr replaced by 0, which is the maximally C compliant approach and avoids using macros apart from _IOLBF which might be #defined to something else.

jbowler:
The MSVC documentation is here:

https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/setvbuf?view=msvc-170

This is not consistent with the ANSI-C documentation (ANSI X3.159-1989, identical to ISO C-90) wihch states in 4.9.5.6 that, "[t]he argument size specifies the size of the array."  So in C-90 size should logically be 0 if the buffer pointer is NULL.  In MSVC the documentation says that a size of less than 2 (i.e. 0 or 1) invokes the "invalid parameter handler" which may well the causing an infinite loop.

The only other possibility is that there is a debug/release msvcrt.dll mismatch on the test system.  That pretty much always causes crashes when the CRT is used from a mismatched debug/release executable or library.  If the change I just pushed doesn't fix the MSVC build I'll just #define the setting out on VC.

jbowler:
Good to go.  I passed 128 as the size parameter to setvbuf on MSVC and 0 for everything else.   Note the description of the (non-standard) setlinebuf function in the Linux and other man pages.  I double checked and setvbuf is giving me 0 as the result on Linux+glibc.

generator/main.cpp
```diff
@@ -135,11 +154,13 @@ int main(int argc, char *argv[])
     printf("Please wait while source files are being generated...\n");
 
     printf("Parsing typesystem file [%s]\n", qPrintable(typesystemFileName));
+    ReportHandler::setContext("Typesytem");
```
mrbean-bremen, line 157: Typesytem -> Typesystem (missed that)

generator/main.cpp
```diff
@@ -53,6 +55,23 @@ void displayHelp(GeneratorSet *generatorSet);
...
+     * interleaving of the stdout and stderr streams when the output is piped
+     * to a file.  Ignore errors here (i.e. ignore closed stdout).
+     *
+     * This causes the check in test with win64_msvc2017_64 to fail.  The
+     * generator crashes here.  The cause is not known, it may be that the
+     * MSVC requires size != 0 even with a NULL buffer (ANSI-C is clear that
+     * the size is the size of the array).
+     */
+#   ifndef Q_CC_MSVC
+#       define _bufsize 0
```
mrbean-bremen, line 68: In the future, we should probably use definitions not dependent on qmake (not now though, they are used elsewhere).

generator/main.cpp
```diff
@@ -53,6 +55,23 @@ void displayHelp(GeneratorSet *generatorSet);
...
+#   else
+#       define _bufsize 128
+#   endif
+    (void)setvbuf(stdout, NULL, _IOLBF, _bufsize);
```
mrbean-bremen, line 72: From the MS documentation for `setvbuf`:
> size
Buffer size in bytes. Allowable range: 2 <= size <= INT_MAX (2147483647). Internally, the value supplied for size is rounded down to the nearest multiple of 2.

So maybe just use a small value (like 2 or 8) for all systems.

mrbean-bremen: > This is not consistent with the ANSI-C documentation (ANSI X3.159-1989, identical to ISO C-90) wihch states in 4.9.5.6 that, "[t]he argument size specifies the size of the array." So in C-90 size should logically be 0 if the buffer pointer is NULL.

From cppstandard:
> If buffer is a null pointer, resizes the internal buffer to size. 

So I would say passing a size is fine with the buffer being NULL. Also, as we already use line buffering instead of full buffering, we could always use a sensible buffer size like 128 instead of basically using no buffering at all.

jbowler: CPP and C standards have diverged, so it's not a safe API to use.

generator/main.cpp
jbowler: I just copied the usage in the rest of the code.  Anyway, it's gone now.

generator/main.cpp
jbowler: Fixed

generator/main.cpp
mrbean-bremen: > CPP and C standards have diverged, so it's not a safe API to use.

The quote from cppstandard explicitely references C99, which I think we can expect to be implemented nowadays (as opposed to C90). I didn't check the standard itself, but I would expect that it is harmonized with the C++ standard as usual.

jbowler: I only have ANSI-C89 (the one I referenced) and POSIX 1003.1 which seems to defer to C90.  I found this for C99:

**Description**

[2](https://port70.net/~nsz/c/c11/n1570.html#7.21.5.6p2) The setvbuf function may be used only after the stream pointed to by stream has been associated with an open file and before any other operation (other than an unsuccessful call to setvbuf) is performed on the stream. The argument mode determines how stream will be buffered, as follows: _IOFBF causes input/output to be fully buffered; _IOLBF causes input/output to be line buffered; _IONBF causes input/output to be unbuffered. If buf is not a null pointer, the array it points to may be used instead of a buffer allocated by the setvbuf function[273)](https://port70.net/~nsz/c/c11/n1570.html#note273) and the argument size specifies the size of the array; _otherwise, size **may** determine the size of a buffer allocated by the setvbuf function_. The contents of the array at any time are indeterminate.

I suspect the ISO C++ standard will say the same thing and cppreference (which can't quote the standard verbatim) is descriptive, not authoritative.  Given that MSVC crashes, and crashes in a CVEable fashion, when it should just accept '0' I don't think setvbuf is safe for final release whereas the other changes are.

mrbean-bremen: You are right, that ("may determine the size") is indeed not as clear as I would have liked. Thanks for digging this up! I will merge it as is now.

PR #132, opened 2023-10-16
generator: AbstractMetaClassList: correct sorting

jbowler:
In some places the list was sorted by pointer (resulting in a random sort) in other cases by name, sometimes with stable_sort, sometimes not. Now sort by name and use a stable_sort.

### DIFF
--- a/generator/abstractmetabuilder.cpp
+++ b/generator/abstractmetabuilder.cpp
@@ -39,8 +39,6 @@
 **
 ****************************************************************************/
 
-#include <algorithm> // for std::sort
-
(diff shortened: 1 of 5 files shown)


jbowler:
This may be part of the reason the WARNING messages output by pythonqt_generator are in a random order; successive runs of the generator produces identical messages in a different order, making checking for errors introduced by a change very difficult.  Unfortunately it isn't the whole reason; the warnings still come out in a random order even with this change.

mrbean-bremen (approved):
👍🏼

PR #134, opened 2023-10-27
Make the generator compile with C++17

mrbean-bremen:
- remove unneeded inheritance from binary_function
- replace auto_ptr with unique_ptr (do not use make_unique, which is not in C++11)


mrbean-bremen:
Going to merge this as these are quite trivial changes.

PR #135, opened 2023-10-27
Some parser fixes

mrbean-bremen:
- add parsing of noexcept exception specifier
- add support for parsing using declaration
- fix error in parsing function declaration if cv qualifiers appear before inline etc

### DIFF
--- a/generator/parser/lexer.cpp
+++ b/generator/parser/lexer.cpp
@@ -1577,6 +1577,20 @@ void Lexer::scanKeyword8()
         }
       break;
 
+    case 'n':
+      if (*(cursor + 1) == 'o' &&
+          *(cursor + 2) == 'e' &&
+          *(cursor + 3) == 'x' &&
+          *(cursor + 4) == 'c' &&
+          *(cursor + 5) == 'e' &&
+          *(cursor + 6) == 'p' &&
+          *(cursor + 7) == 't')
+        {
+          token_stream[(int) index++].kind = Token_noexcept;
+          return;
+        }
+      break;
+
     case 'o':
       if (*(cursor + 1) == 'p' &&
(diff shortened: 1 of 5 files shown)


mrbean-bremen:
This also fixes some of the problems in Qt 5.15. Here are number of generated classes in the CI for the Ubuntu jobs before and after the PR:

|CI|Qt version|qt6|PR|
|Ubuntu 20.04|5.12.8|820|820|
|Ubuntu 22.04|5.15.3|708|810|

Note that the numbers are the same on the master and qt6 branches.

mrbean-bremen:
@jbowler - given [your comment](https://github.com/MeVisLab/pythonqt/pull/136#issuecomment-1783918099) in the other PR I think it would make sense to merge this into main, as it also helps with 5.15 - what do you think?

I intend to further look into generator problems with both Qt5.15 and Qt6. There are several problems with templates (such as variadic template parameters or template default parameters) which are not really relevant for the generated code, as these are mostly ignored, but that cause the generation of classes to fail due to some templated member functions. These should at least be parsed and ignored.

jbowler:
Support for the API extractor helps; when I tested this I could see that quite a few more definitions were being added.  In general I think the changes that recognize new constructs or fix parsing of existing ones should be in the Qt5/master branch.  A good example is if the QTimer function templates could be handled - the thing the QT_DOC #define works round could be removed for Qt5.12.

mrbean-bremen:
Thanks. I could just merge this one if you approve it and cherry pick the changes to master. Next time I would make a PR directly against master.

jbowler:
> Thanks. I could just merge this one if you approve it and cherry pick the changes to master. Next time I would make a PR directly against master.

Yes, certainly approved.

PR #136, opened 2023-10-28
Enum class support

YuriUfimtsev:
Support for parsing enum class declarations (from c++ 11)

### DIFF
--- a/generator/parser/parser.cpp
+++ b/generator/parser/parser.cpp
@@ -1389,8 +1389,24 @@ bool Parser::parseEnumSpecifier(TypeSpecifierAST *&node)
 
   CHECK(Token_enum);
 
+  if (token_stream.lookAhead() == Token_class)
+    {
+      token_stream.nextToken();
+    }
+
   NameAST *name = 0;
   parseName(name);
+
+  if(token_stream.lookAhead() == ':')
+    {
+      token_stream.nextToken();
+      TypeSpecifierAST *ast = 0;
+      if (!parseSimpleTypeSpecifier(ast))
+        {
+          token_stream.rewind((int) start);
+          return false;
+        }
+    }
 
   if (token_stream.lookAhead() != '{')
     {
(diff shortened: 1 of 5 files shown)


mrbean-bremen (approved):
Looks good to me, thanks! @usiems or @jbowler, please have another look.

YuriUfimtsev:
This is done in the purpose of correct generator build and work from Qt 5.12 to Qt 5.15 sources.
If you are welcome to the generated_cpp_512, I will load it throw another PR with "common.prf" changes to support using Qt 5.12 wrappers

jbowler:
The implication is that Qt 5.12 has enum : type style declarations/definitions, i.e. things of the form:

> enum masks : uint32_t {
>    maskTop = 0xff000000,
> 

etc.  I know Qt6 is using those, I couldn't get the generator to work with Qt5.15 but it does work with Qt5.12.  I assume this PR improves the output with Qt5.12 by not skipping the unparseable enums.  Does it remove any of the "something horrible is there" emissions from the generator?  There are several of those and quite a lot of more clear warnings about unparseable things.

So long as the CI checks continue to pass I don't see any issue with any changes to the generator; it simply doesn't work beyond Qt5.12 and I suspect it never will so the value is to people using 5.12 or earlier who wish to add support for currently unsupported Qt subsystems.

YuriUfimtsev:
> Does it remove any of the "something horrible is there" emissions from the generator?
Yes, it is important to notice that this changes were made in order to remove the generator's "SHIT" messages that occur during building generator. Now only warnings (without "shits") are produced by generator while building with qt 5.12 sources.

jbowler:
>Now only warnings (without "shits") are being produced by generator while building with qt 5.12 sources.

There are two in 5.14.2, 5.15.2 and 5.15.11; "ForwardIterator" and "AssociativeIteratorHasKeyAndValue<Iterator,..."  Maybe this explains why QPolygon cbegin/cend are ending up returning QChar*.  QRegion (cbegin and cend) return QRect*, which is maybe reasonable but why on earth support these iterators at all (so far as I can see it's only QPolygon and QRegion, not, for example, QString).

I tried to add <rejection> entries for both of them but this seemed to have no effected on the generated files.

mrbean-bremen:
> Yes, it is important to notice that these changes have been made in order to remove the generator's "SHIT" messages that was occured during building generator.

Thanks for that - this _has_ been a bit embarrasing 😁 

> There are two in 5.14.2, 5.15.2 and 5.15.11; "ForwardIterator" and "AssociativeIteratorHasKeyAndValue<Iterator,..."

That is much better than before. Locally, we also check the parser warnings, of which there are a lot, and probably not all will be fixed. Maybe we should add an option to also log these.

PR #140, opened 2023-11-02
TEST ONLY: this adds the PR139 generated files to the master branch

jbowler:
This is for checking the suitability of the new generated_cpp_ diretories for use with the "master" branch.


jbowler:
Yes, this mirrors my results; everything looks ok.

generated_cpp_514 has a number of instances where cbegin/cend methods which were NOT in the _513 generated files have appeared and return QChar*.  I looked at the differences between qpolygon.h version from 5.13 and 5.14; the only change was Q_DECL_NOTHROW declarations being added as an explicit **noexcept** and no longer protected by Q_COMPILER_RVALUE_REFS.  It's not obvious to me why the declarations were getting dropped before.  Hand editing the return type in the generated files works to an extent.

The relevant change in the log files appears to be from MetaJavaBuilder, in 5.13 it states **"skipping function 'QVector::cbegin', unmatched return type 'const_iterator'"**.  In 5.14 the messages aren't there, however there are messages for "reverse_iterator" and so on.  It looks like somehow const_iterator is now being recognized leading to generation of the bogus wrappers.

jbowler:
Needs more work for Qt5.14; PR141 fixes the initial problems but apparently the generator is now trying to make copy constructors for classes that can't be copied (QSignalBlocker for a start.)  I can't see a way of fixing that, I don't have enough understanding of the typesystem.

YuriUfimtsev:
This PR looks good, CI passed. Are there any problems that prevent it from being merged into master?
 
I have checked PythonQt 5.14 and 5.15 build step with generated_cpp_513 from this PR (https://github.com/YuriUfimtsev/pythonqt/pull/5). CI also passed.
It seems like there are no problems with merging these generated_cpp into master. Actually, it would be great, because newly generated wrappers are needed for the comparison generator step (comparing actually generated wrappers with the version from master), which I'm going to add in the near future.

mrbean-bremen:
@YuriUfimtsev - first, I wouldn't like to merge code generated from another branch into main, though I'm hoping that that branch can be merged soon. 

Second, as I wrote elsewhere - I think it would be better to use the generated code to build PythonQt, e.g. first execute the generator and afterwards compile PythonQt, which would find the generated code. The checked in sources are only used as a fallback. I have used this in a branch to check if the code currently generated for Qt6 compiles (it doesn't yet).

As this would only use the build-in part of the generated code, it would still be helpful to have an additional build step that builds _all_ the generated code to make sure that it at least compiles. Though that would be the next step...

mrbean-bremen:
> As this would only use the build-in part of the generated code, it would still be helpful to have an additional build step that builds all the generated code to make sure that it at least compiles. Though that would be the next step...

I recently noticed that this is not correct - it actually uses all compiled code in the `PythonQt_QtAll` project, which is also built, so the current CI state is indeed sufficient.

mrbean-bremen:
@jbowler: I think this can be revived with the current master branch. I'm not sure if the pre-generated directories are needed for each version (we have switched to generate them ourselves on demand, now that the generator is working more reliably). If they are needed, we might also add directories for 5.15 and 6.5, and make a release in the near future that includes them.

mrbean-bremen:
@jbowler - I think now is the time to add these adapted sources (and add 5.15 and 6.5), as I'm about to make a pre-release for 5.15/6 support. Let us know if you are going to finish this, otherwise we can do this ourselves of course.

jbowler:
Any generated files in the distribution need to have been built in a known way from _known_ Qt sources; the additions have to be made by a core PythonQt maintainer, not me.  If nothing else there is going to be an on-going maintenance requirement imposed by upstream (Qt) security patches.

The generated_cpp files in this PR should certainly not be added.  Any checked in generated_cpp needs to come from an _approved_ Qt source with a known license (and a license that is consistent with that used by PythonQt).

I also agree with other comments that under normal circumstances anyone who builds PythonQt should end up running the generator to get generated_cpp, so normally the files should be generated then.  This would mean a "private" build would use the header files of the private system, probably a reduced set, and a "system" build of a binary for an operating system distribution would use the set supported by the OSD, a set that may be incomplete as a result of incomplete OSD development.

In both cases the license would be appropriate; a commercial user would use the Qt commercial release etc.  Otherwise the result may end up accidentally GPLed or accidentally subject to a Qt commercial license.

So I feel this PR should be closed as, now, completed and any updates and additions regenerated under carefully controlled conditions.

mrbean-bremen:
Thanks, I see your point. As I wrote, we don't need the generated files ourselves, as I changed our build process to create them from the Qt sources we use by the generator. I actually don't know how other people are using the library - so maybe we just wait until somebody chimes in who actually needs them.

As for the license: the checked in sources have always been created from the version of Qt (now) accessible via GitHub, and the license should be compatible with the LGPL we that is used for PythonQt. As the sources are generated from the headers only, I doubt that there would even be a difference if creating them from a closed source version (e.g. > 5.15.11 at the moment). I don't see any problem here, but of course I'm not a lawyer, and I agree that the maintainers of PythonQt are responsible for that.

@florianlink may comment on this, as he is the one who added the generated sources in the first place.

jbowler:
Exactly.  We are in complete agreement insofar as that is ever possible.

PR #142, opened 2023-11-02
generator: fix iterator issues with Qt5.14 headers

jbowler:
const_iterators and other iterators became visible to the generator with changes in Qt5.14.  This removes the corresponding iterators from the affected classes.


mrbean-bremen (approved):
👍🏼

PR #145, opened 2023-11-06
Improve debugging

usiems:
Add some code that helps with debugging the problems that the generator has with newer Qt versions:

- command line option for printing parser errors
- keep track of the current (approximate) source code location, and make current token readable in debugger (only in Debug mode)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -111,5 +111,6 @@ generate.bat
 generator/release
 
 /.vs/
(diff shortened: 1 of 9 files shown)


mrbean-bremen (approved):

mrbean-bremen:
Merging this, as it is helpful for debugging.

PR #146, opened 2023-11-08
Make generator work for Qt 5.15

usiems:
Assorted fixes to make the generator work for Qt 5.15. It's probably best to review the commits individually.

Most of the work consisted of debugging the code paths for classes that didn't come out right and then finding a fix that works. Since I only understand part of the generator (e.g. handling of template parameters still eludes me in parts), I tried to stick to local fixes.


generator/parser/parser.cpp
```diff
@@ -3402,13 +3442,76 @@ bool Parser::parseDeclarationInternal(DeclarationAST *&node)
   return false;
 }
 
-bool Parser::skipFunctionBody(StatementAST *&)
+void popStackUntil(std::vector<int>& stack, int value)
```
mrbean-bremen, line 3445: As can be seen in the CI logs, `#include <vector>` is missing.

mrbean-bremen:
@jbowler - maybe you can have a look. 
In Qt 5.15 (Ubuntu 2204 build) this generates the same number of classes (820) as Qt Qt 5.12 in Ubuntu 2004 (which at the moment does not compile due to a missing include).
I would say that after this PR is done, @YuriUfimtsev has added the check for compiling the generated code, and the regenerated code is checked in (including Qt 5.15 and maybe 5.14), we can make a first release for Qt 5.15.
Qt 6 is still a bit of work...

generator/parser/parser.cpp
```diff
@@ -2300,12 +2300,12 @@ bool Parser::parseBaseClause(BaseClauseAST *&node)
...
 }
 
-bool Parser::parseInitializer(InitializerAST *&node)
+bool Parser::parseInitializer(InitializerAST *&node, bool allowNewStyle)
 {
   std::size_t start = token_stream.cursor();
 
   int tk = token_stream.lookAhead();
-  if (tk != '=' && tk != '(')
+  if (tk != '=' && tk != '(' && !(allowNewStyle && tk == '{'))
```
mrbean-bremen, line 2308: I think `&& (!allowNewStyle || tk == '{')` would be better readable, but that may be just me.

generator/parser/tokens.cpp
```diff
@@ -152,7 +152,9 @@ static char const * const _S_token_names[] = {
   "xor",
   "xor_eq",
   "Q_ENUMS",
-  "Q_ENUM"
+  "Q_ENUM",
+  "Q_INVOKABLE",
+  "<end>"
```
mrbean-bremen, line 157: Is the `<end>` actually needed, or just there as a sentinel?

jbowler:
The generator made with Qt5.15.11LTS fails (aborts) with the **official** Qt downloads for 6.5.3 and 6.6.0:

`Unrecognized character in lexer: .`

(That's a period character.)  That's the last line before the program aborts.  Despite this it succeeds with my Gentoo-patched variants of the same Qt versions.  The line before the above line is a MetaJavaBuilder "WARNING" about:

`enum 'QtWebEngineSettings::UnknownUrlSchemePolicy' does not have a type entry or is not an enum`

(I'm copying these lines by hand, so there might be small typos).  Since these are the official Qt installations (for AMD64 GNU-Linux) it _should_ be possible to repro them anywhere.

mrbean-bremen:
> The generator made with Qt5.15.11LTS fails (aborts) with the official Qt downloads for 6.5.3 and 6.6.0

Yes, I also found this. I haven't looked yet what the actual problem is, but the problematic template code is in `qqmlprivate.h`:
```cpp
    template<class T>
    struct QmlResolved<T, std::void_t<typename T::QmlForeignType>>
    {
        using Type = typename std::conditional<
                QmlTypeHasMarker<T, decltype(&T::qt_qmlMarker_foreign)>::value,
                typename T::QmlForeignType, T>::type;
    };
```
(after commenting this out, the parser runs through), probably somewhere where that template is instantiated.

mrbean-bremen:
I found that after I apply a couple of patches to the generator that I made before, the generator also passes (that's under Windows), though I'm not sure yet what exactly is the problem.

jbowler:
> I found that after I apply a couple of patches to the generator that I made before, the generator also passes (that's under Windows), though I'm not sure yet what exactly is the problem.

It does need a patch; the qqmlprivate.h files are identical in the Gentoo copy of the headers that succeeds and the Qt downloaad that fails.

It sounds like a "use-after-free" and I've seen failures before of that type; a very minor change to the code fixes the problem as does a minor change in the headers.  There is a potential use-after-free in the topo sort which should be removed on principal but that might not be the issue.

jbowler:
I'll try the fix (well, it was a rewrite) that I came up with.

mrbean-bremen:
The problem is the ellipsis inside a type decration that is not expected. I added my commits to make it work, and @usiems will fix the actual type parser  to allow for an ellipsis.

usiems:
The crash is not a real crash, but a qFatal in typeparser.cpp, because Scanner.nextToken() chokes on '.', which didn't occur in any types in Qt 5.15, but now appears in templates types because of ellipsis ("...").

So we have to correctly parse the ellipsis in templates to fix this.

jbowler:
I can't see an ellipsis in that template; the only one outside comments in the whole file (either version) is in the template for `struct StaticCastSelector` and it is in both my versions of the 6.6.0 headers (qqmlprivate.h is identical in both).  My change to the topo sort does not change anything; the output of the generator with and without the change is identical for all the Qt versions I check.  Clearly the sort order problem is somewhere else; the output order does not depend on AbstractMetaClass sortiing.

I'll submit the fix anyway, but it's a rewrite of the function.

generator/parser/parser.cpp
usiems: I thought about that, but found that for me the condition is better readable that way.
By the way, it would be `&& (!allowNewStyle || tk != '{')`

generator/parser/parser.cpp
```diff
@@ -3420,13 +3421,76 @@ bool Parser::parseDeclarationInternal(DeclarationAST *&node)
...
+{
+  std::size_t start = token_stream.cursor();
+  std::vector<int> braceStack;
+  if (token_stream.lookAhead() != '{')
+    {
+      reportError("'{' expected (block start)");
+      return false;
+    }
+  braceStack.push_back('{');
+  nextToken();
```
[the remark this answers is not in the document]
usiems: Ok.

generator/parser/parser.cpp
```diff
@@ -3420,13 +3421,76 @@ bool Parser::parseDeclarationInternal(DeclarationAST *&node)
...
+  while (!braceStack.empty())
+    {
+      int tk = token_stream.lookAhead();
+      switch (tk)
+        {
+        // handle opening braces:
+        case '{':
+        case '[':
+        case '(':
```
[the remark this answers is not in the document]
usiems: It feels a bit uncomfortable for me to have no other checks in place, but ok. The code gets a lot simpler this way.

generator/parser/tokens.cpp
usiems: Just there as a sentinel.

jbowler:
With those commits the crash(qFatal?) against 6.6.0 is fixed however 6.5.3 is crashing with either qFatal (I assume) or a SEGV.  Here's the output:

```
========== Qt: 6.5.3 ==============
WARNING(Typesystem) :: Duplicate type entry: 'QUrl::UrlFormattingOption'
WARNING(Typesystem) :: Duplicate type entry: 'QMultimedia'
QTDIR environment variable not set. This may cause problems with finding the necessary include files.
Please wait while source files are being generated...
Parsing typesystem file [:/trolltech/generator/build_all.txt]
PreProcessing - Generate [.preprocessed.tmp] using [:/trolltech/generator/qtscript_masterinclude.h] and include-paths [/home/jbowler/src/pythonqt/qtinc/6.5.3]
** WARNING expected ``)'' = 40 at /home/jbowler/src/pythonqt/qtinc/6.5.3/QtCore/qnumeric.h:340
corrupted size vs. prev_size
========== Qt: gentoo-6.5.3 ==============
WARNING(Typesystem) :: Duplicate type entry: 'QUrl::UrlFormattingOption'
WARNING(Typesystem) :: Duplicate type entry: 'QMultimedia'
QTDIR environment variable not set. This may cause problems with finding the necessary include files.
Please wait while source files are being generated...
Parsing typesystem file [:/trolltech/generator/build_all.txt]
PreProcessing - Generate [.preprocessed.tmp] using [:/trolltech/generator/qtscript_masterinclude.h] and include-paths [/home/jbowler/src/pythonqt/qtinc/gentoo-6.5.3]
** WARNING expected ``)'' = 40 at /home/jbowler/src/pythonqt/qtinc/gentoo-6.5.3/QtCore/qnumeric.h:340
```

generator/parser/parser.cpp
mrbean-bremen: True of course 😀 And as I wrote, a matter of taste...

generator/parser/tokens.cpp
mrbean-bremen: ...and gone again after the rebase.

mrbean-bremen:
Maybe it's time to setup some Qt6 CI tests, at least in a branch, to catch problems with different versions...

mrbean-bremen (approved):

jbowler:
Here's the traceback of the crash.  This is building the generator with gcc on Linux (gentoo) against Qt5.15.11LTS and then running it against a download of Qt6.5.3 from Qt (i.e. the official free download):

```
gdb pythonqt_generator
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib64/libthread_db.so.1".
Please wait while source files are being generated...
Parsing typesystem file [:/trolltech/generator/build_all.txt]
WARNING(Typesystem) :: Duplicate type entry: 'QUrl::UrlFormattingOption'
WARNING(Typesystem) :: Duplicate type entry: 'QMultimedia'
PreProcessing - Generate [.preprocessed.tmp] using [:/trolltech/generator/qtscript_masterinclude.h] and include-paths [/home/jbowler/src/pythonqt/qtinc/6.5.3]
QTDIR environment variable not set. This may cause problems with finding the necessary include files.
** WARNING expected ``)'' = 40 at /home/jbowler/src/pythonqt/qtinc/6.5.3/QtCore/qnumeric.h:340
Building model using [.preprocessed.tmp]
corrupted double-linked list (not small)
Program received signal SIGABRT, Aborted.
0x00007ffff6efdeac in ?? () from /usr/lib64/libc.so.6
(gdb) bt
#0  0x00007ffff6efdeac in ?? () from /usr/lib64/libc.so.6
#1  0x00007ffff6eaf5c2 in raise () from /usr/lib64/libc.so.6
#2  0x00007ffff6e984ed in abort () from /usr/lib64/libc.so.6
#3  0x00007ffff6e9955c in ?? () from /usr/lib64/libc.so.6
#4  0x00007ffff6f07815 in ?? () from /usr/lib64/libc.so.6
#5  0x00007ffff6f0819b in ?? () from /usr/lib64/libc.so.6
#6  0x00007ffff6f0b003 in ?? () from /usr/lib64/libc.so.6
#7  0x00007ffff6f0bd2a in malloc () from /usr/lib64/libc.so.6
#8  0x00007ffff752fc58 in QHashData::allocateNode (this=<optimized out>,
    nodeAlign=<optimized out>)
    at /usr/src/debug/dev-qt/qtcore-5.15.11-r1/qtbase-everywhere-src-5.15.11/src/corelib/tools/qhash.cpp:479
#9  0x0000555555576fb3 in NameTable::findOrInsert(char const*, unsigned long)
    ()
#10 0x0000555555576d0e in Lexer::scan_identifier_or_keyword() ()
#11 0x00005555555761a9 in Lexer::tokenize(char const*, unsigned long) ()
#12 0x0000555555584905 in Parser::parse(char const*, unsigned long, pool*) ()
#13 0x00005555555ef846 in AbstractMetaBuilder::build() ()
#14 0x00005555555729c5 in main ()
```

Contrary to my previous report the change in https://github.com/MeVisLab/pythonqt/pull/148 seems to make this go away.

From the output the crash appears to be below `GeneratorSetQtScript::buildModel`; the generator is compiled -O2 and that function must be doing tail continuation to `builder.build()` etc...  This is apparently _before_ the call to the changed code in PR148 (it gets call immediately after this) so it's probably not fixed by the PR, just hidden.

mrbean-bremen:
> This is apparently before the call to the changed code in PR148 (it gets call immediately after this) so it's probably not fixed by the PR, just hidden.

Yes, I agree. I doubt that the problem has been introduced with this PR, but it certainly something to be checked and fixed.

mrbean-bremen:
@jbowler - I think that the code is now in a good shape to be merged (for Qt 5). A few more crashes have been fixed in the latest commits, the code generation is now stable, and the code generated for Qt5.15 looks good.

If you agree, I will merge this, and further changes related to Qt6 will go into separate PRs. This then can be used as base for newly generated source, for Qt <= Qt5.15. The new CI checks worked on by @YuriUfimtsev will also help to verify this.

jbowler:
> @jbowler - I think that the code is now in a good shape to be merged (for Qt 5).

I've had the 6.5.3 crash sitting under a debugger for almost a week.  I'll try the test again with the latest changes.  I'm pretty certain it isn't 6.5.3 specific.  It occurs in the lexical analysis  somewhere low down in a QHash update but it happens with CONFIG+=DEBUG as well so it is very unlikely to be a memory overwrite.

I think this needs to have been fixed before the merge.  Anyway, I'll test again now there have been substantial changes.

jbowler:
FWIW I've just discovered that the original crash disappears if I `git revert 14bbc092e5f37c8ba53721ae18603332b22227fe`  This makes sense because the crash was happening while parsing a template for an istream class ">>" operator.  I've yet to test the latest changes.

jbowler:
With the latest HEAD, i.e. 2aecabb6c961a6c4a51f0ed9817945caec3b756a all my generation tests succeed.  This just means that the generator works without explicitly crashing; I don't know whether the generated_cpp is valid/usable!

The change that originated the crash is now:

dd8f906e2f4826e49cb9d1bc8f78aab89caa6200

I note that the last fix (the HEAD, as referenced above) fixes addition of two tokens at once in the lexer and the above commit does, indeed, change the code to add two rather than one token for >>.  The crash is when the lexical analysis cursor is precisely here:

`template <typename Stream, typename, typename = void>
struct has_istream_operator : std::false_type {};
template <typename Stream, typename T>
struct has_istream_operator<Stream, T, std::void_t<decltype(detail::reference<Stream>() >> `**d**`etail::reference<T>())>>
: std::true_type {};
`

This explains the crash to me enough for me to be happy with a merge back to the main line.

mrbean-bremen:
Thank you - merging then!

jbowler:
Ok, but did anyone try the headers generated by a Qt5.15.11 build using the official Qt5.11.3 header files?  When I try that I get this:

`g++ -L/usr/lib64 -ldl -lm -Wl,-O1 -Wl,-rpath,/usr/lib64 -Wl,-rpath-link,/usr/lib64 -shared -Wl,--no-undefined -Wl,-soname,libPythonQt_QtAll-Qt5-Python3.11.so.3 -o libPythonQt_QtAll-Qt5-Python3.11.so.3.2.0 PythonQt_QtAll.o com_trolltech_qt_core0.o com_trolltech_qt_core1.o com_trolltech_qt_core2.o com_trolltech_qt_core_init.o com_trolltech_qt_gui0.o com_trolltech_qt_gui1.o com_trolltech_qt_gui2.o com_trolltech_qt_gui3.o com_trolltech_qt_gui4.o com_trolltech_qt_gui5.o com_trolltech_qt_gui_init.o com_trolltech_qt_svg0.o com_trolltech_qt_svg_init.o com_trolltech_qt_sql0.o com_trolltech_qt_sql_init.o com_trolltech_qt_network0.o com_trolltech_qt_network1.o com_trolltech_qt_network_init.o com_trolltech_qt_opengl0.o com_trolltech_qt_opengl_init.o com_trolltech_qt_xmlpatterns0.o com_trolltech_qt_xmlpatterns_init.o com_trolltech_qt_multimedia0.o com_trolltech_qt_multimedia1.o com_trolltech_qt_multimedia2.o com_trolltech_qt_multimedia3.o com_trolltech_qt_multimedia_init.o com_trolltech_qt_uitools0.o com_trolltech_qt_uitools_init.o moc_com_trolltech_qt_core0.o moc_com_trolltech_qt_core1.o moc_com_trolltech_qt_core2.o moc_com_trolltech_qt_gui0.o moc_com_trolltech_qt_gui1.o moc_com_trolltech_qt_gui2.o moc_com_trolltech_qt_gui3.o moc_com_trolltech_qt_gui4.o moc_com_trolltech_qt_gui5.o moc_com_trolltech_qt_svg0.o moc_com_trolltech_qt_sql0.o moc_com_trolltech_qt_network0.o moc_com_trolltech_qt_network1.o moc_com_trolltech_qt_opengl0.o moc_com_trolltech_qt_xmlpatterns0.o moc_com_trolltech_qt_multimedia0.o moc_com_trolltech_qt_multimedia1.o moc_com_trolltech_qt_multimedia2.o moc_com_trolltech_qt_multimedia3.o moc_com_trolltech_qt_uitools0.o  -lpython3.11 -ldl -lm -L../../lib/../lib -lPythonQt-Qt5-Python3.11 /usr/lib64/libQt5PrintSupport.so /usr/lib64/libQt5Svg.so /usr/lib64/libQt5OpenGL.so /usr/lib64/libQt5MultimediaWidgets.so /usr/lib64/libQt5QuickWidgets.so /usr/lib64/libQt5UiTools.a /usr/lib64/libQt5Widgets.so /usr/lib64/libQt5Multimedia.so /usr/lib64/libQt5Quick.so /usr/lib64/libQt5Gui.so /usr/lib64/libQt5Sql.so /usr/lib64/libQt5XmlPatterns.so /usr/lib64/libQt5QmlModels.so /usr/lib64/libQt5Qml.so /usr/lib64/libQt5Network.so /usr/lib64/libQt5Xml.so /usr/lib64/libQt5Core.so -lGL -pthread    
/usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../x86_64-pc-linux-gnu/bin/ld: PythonQt_QtAll.o: in function `PythonQt_QtAll::init()':
PythonQt_QtAll.cpp:(.text+0x3f): undefined reference to `PythonQt_init_QtQml(_object*)' 
/usr/lib/gcc/x86_64-pc-linux-gnu/13/../../../../x86_64-pc-linux-gnu/bin/ld: PythonQt_QtAll.cpp:(.text+0x46): undefined reference to `PythonQt_init_QtQuick(_object*)'
collect2: error: ld returned 1 exit status
`

The build with the 5.15.11 header files installed by Gentoo succeeds, but maybe I don't have somelike (like Qml) installed; I don't use Qml.

jbowler:
I think it may just be a feature of how I had to set up the official 5.15.11 headers; Qt5.15.2 succeeds, but it would be a good idea for someone else with *official* 5.15.11 headers to check this out.

PR #147, opened 2023-11-09
Improve CI with sanitizers

YuriUfimtsev:
- Add sanitizers to the PythonQt testing step and the generator step. Due to the large number of leaks, leak_detects is temporarily disabled.
- Add consistency check between TOKEN_KIND enum and _S_token_names array, because otherwise CI fails with the ASan error "Array index out of bounds".


mrbean-bremen (approved):
Looks good - thanks!

PR #148, opened 2023-11-09
Fix: AbstractMetaBuilder::classesTopologicalSorted

jbowler:
This fixes explicit new/delete, storage leaks (missing delete) and iteration over a QHash while modifying it (deleting specific keys in an iterator).

Tested against the current make_generator_work_for_5.15 branch; it generates identical output and crashes in the same places (6.5.3).

### DIFF
--- a/generator/abstractmetabuilder.cpp
+++ b/generator/abstractmetabuilder.cpp
@@ -2480,55 +2480,100 @@ void AbstractMetaBuilder::dumpLog()
 
 AbstractMetaClassList AbstractMetaBuilder::classesTopologicalSorted() const
 {
-    AbstractMetaClassList res;
+    /* This function is the standard topological of a Directed Acyclic Graph
+     * (a DAG).  It outputs a partially ordered list of the nodes in the graph
+     * such that a node is output after all of its children.
+     *
+     * In the previous implementation it accounted for around 68-69% of the
+     * entire run time of pythonqt_generator.  The implementation also leaked
+     * memory and modified a QHash while iterating over it with a QHashIterator.
+     *
+     * This version is somewhat like the python equivalent using list
+     * comprehensions.
+     */
 
-    AbstractMetaClassList classes = m_meta_classes;
-    classes.sort();
+    /* Build a hash list of QSet<class> for each class.  'class' is represented
+     * by a pointer to the AbstractMetaClass from m_meta_classes.
+     */
+    ReportHandler::debugSparse(QString("TSORT: %1 meta classes")
+            .arg(m_meta_classes.count()));
+    QHash<AbstractMetaClass*, QSet<AbstractMetaClass*>> classes;
+    classes.reserve(m_meta_classes.count());
 
-    QSet<AbstractMetaClass*> noDependency;
-    QHash<AbstractMetaClass*, QSet<AbstractMetaClass* >* > hash;
-    for (AbstractMetaClass *cls :  classes) {
-        QSet<AbstractMetaClass* > *depends = new QSet<AbstractMetaClass* >();
-
+    for (auto cls : m_meta_classes) {
+        /* Add the baseClass and the interfaces the class uses.  The latter
+         * are stored in an AbstractMetaClassList which is uses a QList, so:
+         */
+        auto entry(classes.insert(cls, QSet<AbstractMetaClass*>(
+                    cls->interfaces().cbegin(), cls->interfaces().cend())));
         if (cls->baseClass())
-            depends->insert(cls->baseClass());
-
-        for (AbstractMetaClass *interface :  cls->interfaces()) {
-            AbstractMetaClass *impl = interface->primaryInterfaceImplementor();
-            if (impl == cls)
-                continue;
-            depends->insert(impl);
-        }
-
-        if (depends->empty()) {
-            noDependency.insert(cls);
-        } else {
-            hash.insert(cls, depends);
-        }
+            entry.value().insert(cls->baseClass());
+        entry.value().remove(cls); // may come from interfaces
     }
 
-    while (!noDependency.empty()) {
-        for (AbstractMetaClass *cls :  noDependency.values()) {
-            if(!cls->isInterface())
-                res.append(cls);
-            noDependency.remove(cls);
-            QHashIterator<AbstractMetaClass*, QSet<AbstractMetaClass* >* > i(hash);
-            while (i.hasNext()) {
-                i.next();
-                i.value()->remove(cls);
-                if (i.value()->empty()) {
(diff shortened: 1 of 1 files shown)


generator/abstractmetabuilder.cpp
```diff
@@ -2480,55 +2480,100 @@ void AbstractMetaBuilder::dumpLog()
...
+            .arg(m_meta_classes.count()));
+    QHash<AbstractMetaClass*, QSet<AbstractMetaClass*>> classes;
+    classes.reserve(m_meta_classes.count());
+
+    for (auto cls : m_meta_classes) {
+        /* Add the baseClass and the interfaces the class uses.  The latter
+         * are stored in an AbstractMetaClassList which is uses a QList, so:
+         */
+        auto entry(classes.insert(cls, QSet<AbstractMetaClass*>(
+                    cls->interfaces().cbegin(), cls->interfaces().cend())));
```
mrbean-bremen, line 2508: This is not upwards compatible. Using 
```cpp
#if QT_VERSION >= 0x060000
        auto interfaces = QSet<AbstractMetaClass*>(
            cls->interfaces().cbegin(), cls->interfaces().cend());
#else
        auto interfaces = QSet<AbstractMetaClass*>::fromList(cls->interfaces());
#endif
        auto entry(classes.insert(cls, interfaces));
```
instead fixes it. Actually `cBegin()` / `cEnd()` have been introduced somewhere in Qt 5.15, so the check could  be changed, but I just did the same as in other places in the code.

generator/abstractmetabuilder.cpp
```diff
@@ -2480,55 +2480,100 @@ void AbstractMetaBuilder::dumpLog()
...
+
+    /* Build a hash list of QSet<class> for each class.  'class' is represented
+     * by a pointer to the AbstractMetaClass from m_meta_classes.
+     */
+    ReportHandler::debugSparse(QString("TSORT: %1 meta classes")
+            .arg(m_meta_classes.count()));
+    QHash<AbstractMetaClass*, QSet<AbstractMetaClass*>> classes;
+    classes.reserve(m_meta_classes.count());
+
+    for (auto cls : m_meta_classes) {
```
mrbean-bremen, line 2503: You have removed the sorting of `m_meta_classes` which you had introduce earlier for a more stable output. Isn't that needed anymore?

mrbean-bremen (commented):
This looks much cleaner than before, especially as it avoids the `new` / `delete` stuff. I'm not sure about the performance - the numbers seen in the CI are not really conclusive - but I don't really care as much about performance here as about correctness and stability.

generator/abstractmetabuilder.cpp
mrbean-bremen: Ah, just noticed your comment in the other PR - apparently it indeed doesn't matter.

jbowler:
> I'm not sure about the performance.

That's why I started looking at it; the GCC with -pg showed this one function (including the things it calls) as accounting for a very large percentage of the run time (I can't remember the number but it was around half).  After I'd done the changes the % dropped by a factor of 5.  However gprof lies; the actual run-time of the generator was not changed significantly!

I still don't understand where the perf is going but "-p" profiling seems to have been dropped from the GCC (or maybe I don't know where it is) and I don't have anything on my Gentoo system to do sampled profiling (where the PC is sampled regularly then mapped back to the function based on the debug information).  I always found that much better and VS supported it really well (this was 20 years ago).  You could try profiling a run under VS but I don't know which version you are using; good support probably requires Pro and the best support needed the Enterprise version :-(

mrbean-bremen:
As I wrote, I don't really care much about the performance, as the generator will usually be run only once for a version to generate the code (except in the CI, of course), which is then used. Or if the checked in generated code is used, it need not to be run at all.

Having a version that creates correct code and is stable is much more important from my point of view. As I wrote elsewhere, if we can get it to a stage where 5.15 sources are generated reliably, we can newly check in the newly generated code and call it a release. 

What bugs me a bit are the crashes you get on your system with Qt6, but that may not be relevant for the Qt 5.15 generator.

generator/abstractmetabuilder.cpp
```diff
@@ -2496,55 +2496,100 @@ void AbstractMetaBuilder::dumpLog()
...
-
-    AbstractMetaClassList classes = m_meta_classes;
-    classes.sort();
-
-    QSet<AbstractMetaClass*> noDependency;
-    QHash<AbstractMetaClass*, QSet<AbstractMetaClass* >* > hash;
-    for (AbstractMetaClass *cls :  classes) {
-        QSet<AbstractMetaClass* > *depends = new QSet<AbstractMetaClass* >();
-
+    /* This function is the standard topological of a Directed Acyclic Graph
```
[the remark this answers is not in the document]
jbowler: Yes.  Odd, it's line 2483 on my system.

I'll update the comment about the time too since I sort-of disproved it.

generator/abstractmetabuilder.cpp
jbowler: The 'sort' was a side-effect of calling the function; it modified the global m_meta_classes for no reason.  The function itself sorts the output into a new AbstractMetaClassList (this, after all, is the point) and the order within the output is randomized by the use of QSet.

I've already changed the sort function to use function names, not pointers, but the QSets necessarily have to use pointers (in case the function name is duplicated; should be impossible but who knows?)  So the original sort no longer works as, maybe, expected.

The QSet stuff could be fixed too, but it means converting the `handled `QSet to a QList then sorting that immediately before it is appended to `result`.  It might produce more reliable sorted output but I still don't understand where the randomized output comes from; this function is called right at the start of the "`generate`" function.

I don't think sorting the output is a good idea at this point; this change does not alter the final sort order (making it easy to check for errors in the output), sorting would add overhead and I strongly suspect it would still not change the final sort order making the operation (like the original sort) a waste.

jbowler:
Github has a weird interface.  I think I double clicked something.

jbowler:
> What bugs me a bit are the crashes you get on your system with Qt6, but that may not be relevant for the Qt 5.15 generator.

If it crashes, even if it crashes with invalid input, it's a problem.

jbowler:
I must have mis-tested this before; this change **does** make the two crashes I was seeing go away.  I can't be sure if it fixes them of course; they may be intermittent, but with the current make_generator* branch I get crashes but with this change I didn't see them.  Since I can repro the crashes I can probably find out where they are but that may not help if it's heap corruption.  Anyway I can apparently repro with commit 94c8ee2f5f09baa807ace6f297440a6517be61df reliably so it's safe to merge this if the CI checks pass now.

mrbean-bremen (approved):

PR #149, opened 2023-11-11
Add CI for the latest Qt5 and Qt6

iakov:
Example CI workflow for Qt5.15 and Qt6.x builds on Ubuntu-latest worker. 
Small additional change from `make -j 2` to `make -j $(nproc)` is useful for self-hosted and larger Github runners.

### DIFF
--- a/.github/workflows/build_latest.yml
+++ b/.github/workflows/build_latest.yml
@@ -1,0 +1,71 @@
+name: Latest Qt5/6
+
+on:
+  push:
+    branches: 
+      - master
+  pull_request:
+      
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  ubuntu:
+    strategy:
+      fail-fast: false
+      matrix:
+        qt-version: [ '5.*', '6.*']
+        python-version: [ '3.12' ]
+    runs-on: ubuntu-latest
+    steps:
+      
+    - name: Install Qt ${{matrix.qt-version}}
+      uses: jurplel/install-qt-action@v3
+      with:
+        version: ${{ matrix.qt-version }}
+        modules: ${{startsWith(matrix.qt-version, '6') && 'qt5compat' || '' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/build_latest.yml
```diff
@@ -0,0 +1,71 @@
...
+    - name: Set environment
+      run: |
+        QT_VERSION_MAJOR=$(cut -f 1 -d . <<< "${{matrix.qt-version}}")
+        echo "QT_VERSION_MAJOR=$QT_VERSION_MAJOR" >> $GITHUB_ENV
+        QTDIR=$(eval echo "\$Qt${QT_VERSION_MAJOR}_DIR")
+        echo "QTDIR=$QTDIR" >> $GITHUB_ENV
+        echo "$QTDIR/bin" >> $GITHUB_PATH
+        echo "$pythonLocation/bin" >> $GITHUB_PATH
+
+    - name: Build PythonQt
```
mrbean-bremen, line 47: If you want to make this work as is, you probably have to create the generator and generate the code before building PythonQt, as PythonQt needs some of the generated files. 
Or you would have to modify the project file that Qt6 uses the headers generated for Qt5...

mrbean-bremen: @usiems is currently working on the generator to make it work for Qt6 (it already works pretty good for 5.15, at least in that branch), so hopefully this may be possible some time not so far away. Another thing that would help of course would be to actually compile all of the generated sources, not only the part that is needed for PythonQt. I understand that this is something @YuriUfimtsev has on his to-do list?

mrbean-bremen: Sounds good! You probably read my [related comment](https://github.com/MeVisLab/pythonqt/pull/140#issuecomment-1809637776) in response to The Chosen One (will have to call him that now 😁)

iakov:
Obsolete now. Can be closed. The only feature here is the unified workflow file example for both Qt6 and Qt5

Issue #138, opened 2023-11-01
CI tests for generator component are missing


iakov:
Missing test for generator, but definitely some regression tests are required. I suggest implementing the following simple steps.
1. Add ASAN_OPTIONS to generation step, at least `ASAN_OPTIONS="detect_stack_use_after_return=1:fast_unwind_on_malloc=0"`
1. Change each CI job to have an additional step to build with freshly generated sources.
1. (Optional, arguable) add step to diff the `generated_cpp` sources with the currently used `generated_cpp_XXX` to enforce the regular update. 

The last one should be done after the fixes in generator for the known [non-determinism](https://github.com/MeVisLab/pythonqt/pull/133#issuecomment-1783927263).

mrbean-bremen:
@iakov - are you planning to make a PR for this, or should someone else look at it?

iakov:
@YuriUfimtsev is "the chosen one". This is the issue for him, he knows, we've discussed this in person.

iakov:
Probably, `detect_leaks=1` can be turned on (with tiny suppressions for globals) after #148

mrbean-bremen:
> Probably, detect_leaks=1 can be turned on (with tiny suppressions for globals) after https://github.com/MeVisLab/pythonqt/pull/148

Yes, this is the hope... let's wait and see, hopefully that branch will be merged soon.

mrbean-bremen:
I had switched on `detect_leaks=1` in a branch and noticed that there is still a lot of leaks in the generator. Personally, I'm not too worried about this, as the generator is only run on demand every once in a while, and it doesn't bother me if that takes a few MB more of RAM. I don't think it is worth the possibly large effort of fixing this, but if anyone is willing, I won't object 😀 

As for the other other goals: there is currently the "Check generated_cpp" build job that runs for Qt 5.12, 5.15 and 6.5, for ubuntu and Windows (MSVC). We also may add Macos builds here, and if needed more versions. We could also change the existing jobs to compile the generated sources, and consolidated the existing tests. Here is a list of the versions currently tested in CI for reference:

Build
```
ubuntu 20.04  Qt 5.12.8    Python 3.8.10
ubuntu 22.04  Qt 5.15.3    Python 3.10.12
centos 7      Qt 5.9.7     Python 2.7.5
rockylinux 9  Qt 5.15.3    Python 3.9.18
MacOS 11      Qt 5.9.9     Python 3.6.15
[...]

PR #150, opened 2023-11-12
Reduce memory leaks

YuriUfimtsev:
A step towards #138.
Generator behaviour doesn't change (for example, based on wrappers generated for qt 5.12 before and after this fix), but the amount of leaked memory decreases by a third (from 57 to 37 MB). The measurements were carried out on the make_generator_work_for_5.15 branch.
I will try to continue to reduce the number of leaks in the near future.


mrbean-bremen (approved):
Thanks!

PR #151, opened 2023-11-17
Qt5 latest workflow

YuriUfimtsev:
CI workflow for Qt5 latest builds on Ubuntu-latest worker based on #149.
Adds a new step that allows to build PythonQt with freshly generated code.
Now contains qt 5.12 version for testing a new scenario, but easily can be changed to 5.* (equivalent to 5.15) when PR #146 is merged.

### DIFF
--- a/.github/workflows/build_latest.yml
+++ b/.github/workflows/build_latest.yml
@@ -1,0 +1,71 @@
+name: Latest Qt5/6
+
+on:
+  push:
+    branches: 
+      - master
+  pull_request:
+      
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  ubuntu:
+    strategy:
+      fail-fast: false
+      matrix:
+        qt-version: [ '5.*', '6.*']
+        python-version: [ '3.12' ]
+    runs-on: ubuntu-latest
+    steps:
+      
+    - name: Install Qt ${{matrix.qt-version}}
+      uses: jurplel/install-qt-action@v3
+      with:
+        version: ${{ matrix.qt-version }}
+        modules: ${{startsWith(matrix.qt-version, '6') && 'qt5compat' || '' }}
+        archives: 'qtmultimedia qtlinuxextras qtbase qttools icu qtwayland'
+
+    - name: Setup Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v4
+      with:
+        python-version: '${{ matrix.python-version }}'
+
+    - name: Checkout PythonQt
+      uses: actions/checkout@v3
+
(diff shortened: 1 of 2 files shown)


.github/workflows/build_latest.yml
```diff
@@ -0,0 +1,77 @@
+name: Latest Qt5
```
mrbean-bremen, line 1: Maybe change the name to something like "Use generated code", as this is certainly not the latest Qt5. There are two othyer workflows (ubuntu 22.04 and WinGW 7.3) with Qt 5.15.

mrbean-bremen (commented):
This looks good to me, thanks! One question (also to @iakov) - I wonder if it would make sense to use this as the default workflow, e.g. integrate it also into the other workflows?

.github/workflows/build_latest.yml
```diff
@@ -0,0 +1,77 @@
...
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  ubuntu:
+    strategy:
+      fail-fast: false
+      matrix:
+        qt-version: [ '5.12.*' ]
```
mrbean-bremen, line 18: Now that the other branch is merged, you may try to add "5.15.*" to the matrix.

YuriUfimtsev:
> This looks good to me, thanks! One question (also to @iakov) - I wonder if it would make sense to use this as the default workflow, e.g. integrate it also into the other workflows?

After discussion with @iakov,
Sounds good but the main purpose of this workflow is to prevent from occurring some generator problems like string duplication founded in the #137 PR, that breaks the PythonQt build.
So the "mission" of this workflow is to ensure that generator fixes don't break generator and don't negatively affect to the generated_cpp.

We don't think that it is needed for every qt version, moreover, it seems to be unneeded to check freshly generated code for qt less than 5.6 version, because these versions are not the LTS, not supported, and correctly generated_cpp, which are needed for PythonQt, are already in master. For the latest versions, especially for the LTS versions, it seems more reasonable to do a check.

Thus, the task is to check the correctness of the new versions of the generator and the generating on the new versions of Qt, so a separate workflow for this seems to us a good idea

.github/workflows/build_latest.yml
YuriUfimtsev: Yes, there is some misunderstanding between "latest qt" and 5.12 :)
I will rename it to the more suitable form

mrbean-bremen:
Agreed. I still would like to see 5.15 added here, though.

mrbean-bremen (approved):

PR #153, opened 2023-11-17
Make number of classes per generated file configurable

usiems:
This introduces the --max-classes-per-file command line argument. The default for this is 30, as before.

The motivation for this is that comparing the output for different versions (of Qt or the generator) gets easier when everything is one file, so that added or removed classes don't change the output of subsequent files. For this purpose one can use `--max-classes-per-file=10000`.

### DIFF
--- a/generator/generatorsetqtscript.h
+++ b/generator/generatorsetqtscript.h
@@ -62,6 +62,7 @@ public:
 
 private:
     MetaQtScriptBuilder builder;
+    int maxClassesPerFile{30};
   
 };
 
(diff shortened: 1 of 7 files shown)


generator/generatorsetqtscript.h
```diff
@@ -62,6 +62,7 @@ class GeneratorSetQtScript : public GeneratorSet
 
 private:
     MetaQtScriptBuilder builder;
+    int maxClassesPerFile{30};
```
mrbean-bremen, line 65: I wonder if we should use a large number as default instead to make changes by default better comparable. I assume that the restriction had to do with limitations / performance issues with earlier compilers and IDEs, and I doubt that this is still needed nowadays. Would like to hear other opinions...

mrbean-bremen (approved):

mrbean-bremen:
I'm going to merge this as it is needed for the other PR. We can always change the default later.

mrbean-bremen:
Actually it is the other PR which is needed...

PR #154, opened 2023-11-17
Make generator work for Qt6

usiems:
These is my work on making the PythonQt_generator work with Qt6 (I worked with the header files of Qt6.5.1).

I believe there is still room for improvement - there are still a lot of parser errors, - but the generated code looks sensible to me. I didn't try to compile it yet, though.


.github/workflows/build_latest.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        qt-version: [ '5.12.*', '5.15.*' ]
+        qt-version: [ '5.12.*', '5.15.*', '6.5.*' ]
```
mrbean-bremen, line 18: You also have to add the `qt5compat` module in the "Install Qt" action. As it was done in the other PR:
```
with:
  version: ${{ matrix.qt-version }}
  modules: ${{startsWith(matrix.qt-version, '6') && 'qt5compat' || '' }}
  ...
```

githubuser0xFFFF:
@usiems Just want to say thank you and give some feedback. I just managed to generate the bindings for Qt6.5.3 using the `make_generator_work_for_qt6` branch. After generating the files, I managed to successfully compile and use the `PythonQt-Qt6-Python3.8.dll` library in our application using the MinGW 11.20 compiler.

In order to build the library, I needed to manually comment out some code in the generated core and gui builtins but then everything worked - great 👍.

Just in case you are interested, here are the changes I needed to do to the generated files to build the library:

https://github.com/githubuser0xFFFF/pythonqt/commit/e4d8c735bb63b94d0c89492794e1a9d33e0fc356

usiems:
So, the generator is in pretty good shape for Qt 6.5 and 6.6 now. For Qt 6.6 QRunnable and QThreadPool are missing (compared to 6.5), but these were not reasonably usable from Python anyway.

There are still some open issues I would like to do, like handling of QAnyStringView/QByteArrayView parameters, treating Q_PROPERTIES of gadgets as properties, or putting enum class values in their own enum scope, but those will be done in separate issues.

PR #161, opened 2023-12-08
Support QStringView and QAnyStringView as argument and return type

usiems:
QStringView and QAnyStringView are interfaces to treat different string origins the same. In Qt5 we had QStringRef instead.

To support this as an argument from Python, we also need to (temporarily) store the original QString somewhere, and then create the Q(Any)StringView from it. The old code supported QStringRef as return value, but didn't allow it as argument type, as far as I can see. But the code supports QString* as argument type, which is a somewhat similar case. I took my inspiration how to handle the view classes from that.

After fixing this it became apparent that in C++ there often were several overloads of a method that amounted to basically the same call interface in Python (QString, QLatin1String, QStringView). So I wrote code to remove superfluous overloads.

This is untested yet, I asked mrbean-bremen for that, since he is currently working with this code.

### DIFF
--- a/generator/typesystem.cpp
+++ b/generator/typesystem.cpp
@@ -1509,22 +1509,27 @@ TypeDatabase *TypeDatabase::instance()
 
 TypeDatabase::TypeDatabase() : m_suppressWarnings(true)
 {
-    addType(new StringTypeEntry("QString"));
+    StringTypeEntry* mainStringType = new StringTypeEntry("QString");
+    addType(mainStringType);
 
     StringTypeEntry *e = new StringTypeEntry("QLatin1String");
     e->setPreferredConversion(false);
+    e->setEquivalentType(mainStringType);
     addType(e);
 
     e = new StringTypeEntry("QStringRef");
     e->setPreferredConversion(false);
+    e->setEquivalentType(mainStringType);
     addType(e);
 
     e = new StringTypeEntry("QStringView");
     e->setPreferredConversion(false);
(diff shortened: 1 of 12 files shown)


generator/typesystem.cpp
```diff
@@ -1509,14 +1509,27 @@ TypeDatabase *TypeDatabase::instance()
...
+
+    e = new StringTypeEntry("QStringView");
+    e->setPreferredConversion(false);
+    e->setEquivalentType(mainStringType);
+    addType(e);
+
+    e = new StringTypeEntry("QAnyStringView");
+    e->setPreferredConversion(false);
+    e->setEquivalentType(mainStringType);
     addType(e);
```
mrbean-bremen, line 1533: What about `QUtf8String`? And for that matter, it probably also makes sense to add `QByteArrayView`.

mrbean-bremen (commented):
I have tested a few of the newly generated methods with `QStringView` and `QAnyStringView` arguments, and they work fine. `QDate::toString(QString)` is still not generated, but that may have to do with the second optional `QCalender` argument.

generator/typesystem.cpp
usiems: You mean QLatin1StringView and QUtf8StringView? I didn't deem them useful, because in most cases there is also a variant of the same method with QAnyStringView or QString as argument.

QByteArrayView might be useful in some cases, but is a little bit more complicated to handle.

mrbean-bremen (approved):

PR #162, opened 2023-12-11
Add "enum class" values to the enum type in Python

usiems:
Enum class values seem to have pretty generic names (as there is no danger
of name clashes in C++, as they must always be prefixed with the enum
class name). So add them to the enum type in Python too, so one can
always write something like QActionGroup.ExclusionPolicy.None instead of
QActionGroup.None.

Ideally we wouldn't add the enum values to the parent class at all, but
I don't want to break compatibility with older versions of PythonQt.

Also added code to escape names that are reserved in Python.

### DIFF
--- a/src/PythonQtClassInfo.cpp
+++ b/src/PythonQtClassInfo.cpp
@@ -853,12 +853,26 @@ PyObject* PythonQtClassInfo::findEnumWrapper(const QByteArray& name, PythonQtCla
   }
 }
 
+// required for the code below (for versions before Qt 6)
+Q_DECLARE_METATYPE(PythonQtObjectPtr)
+
 void PythonQtClassInfo::createEnumWrappers(const QMetaObject* meta)
 {
   for (int i = meta->enumeratorOffset();i<meta->enumeratorCount();i++) {
     QMetaEnum e = meta->enumerator(i);
     PythonQtObjectPtr p;
     p.setNewRef(PythonQtPrivate::createNewPythonQtEnumWrapper(e.name(), _pythonQtClassWrapper));
+#if QT_VERSION > 0x050800
+    if (e.isScoped()) {
+      // add enum values to the enum type itself, in case enum value names are so generic
+      // that they are not unique
+      for (int j = 0; j < e.keyCount(); j++) {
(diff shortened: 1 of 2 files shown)


src/PythonQtClassInfo.cpp
```diff
@@ -49,6 +49,11 @@
 
 QHash<QByteArray, int> PythonQtMethodInfo::_parameterTypeDict;
 
+// List of words that are reserved in Python, but not in C++, so they need escaping
+QSet<QByteArray> PythonQtClassInfo::_reservedNames{
+  "None", "True", "False"
+};
```
mrbean-bremen, line 55: I guess this is not a full list, just the names that are reasonably can be used in specifiers - otherwise there would be a few missing (like "elif", "with" or "pass").

src/PythonQtClassInfo.cpp
```diff
@@ -859,6 +864,17 @@ void PythonQtClassInfo::createEnumWrappers(const QMetaObject* meta)
...
     PythonQtObjectPtr p;
     p.setNewRef(PythonQtPrivate::createNewPythonQtEnumWrapper(e.name(), _pythonQtClassWrapper));
+#if QT_VERSION > 0x050800
+    if (e.isScoped()) {
+      // add enum values to the enum type itself, in case enum value names are so generic
+      // that they are not unique
+      for (int j = 0; j < e.keyCount(); j++) {
+        PythonQtObjectPtr enumValuePtr;
+        enumValuePtr.setNewRef(PythonQtPrivate::createEnumValueInstance(p.object(), e.value(j)));
+        p.addVariable(escapeReservedNames(e.key(j)), QVariant::fromValue(enumValuePtr));
```
mrbean-bremen, line 874: `QVariant` needs a registered type for  Qt < 6.

src/PythonQtClassInfo.cpp
usiems: Exactly. Only the names that would reasonably occur in an enum value name (because e.g. they start with a upper case letter).

mrbean-bremen (approved):

PR #167, opened 2023-12-12
Replace INCLUDE by PYTHONQT_INCLUDE env var

mrbean-bremen:
- avoids accidental inclusion of system-defined include dirs
- minor documentation updates


generator/main.h
```diff
@@ -55,20 +55,16 @@ struct Preprocess
...
-#endif
+        QChar pathSplitter = QDir::listSeparator();
 
-        // Environment INCLUDE
-        QString includePath = getenv("INCLUDE");
+        // Environment PYTHONQT_INCLUDE
+        QString includePath = getenv("PYTHONQT_INCLUDE");
         if (!includePath.isEmpty())
-            includes += includePath.split(path_splitter);
+            includes += includePath.split(pathSplitter);
```
usiems, line 63: Perhaps add parameter Qt::SkipEmptyParts?

generator/main.h
```diff
@@ -55,20 +55,16 @@ struct Preprocess
...
 
         // Includes from the command line
         if (!commandLineIncludes.isEmpty())
-            includes += commandLineIncludes.split(path_splitter);
+            includes += commandLineIncludes.split(pathSplitter);
```
usiems, line 67: Perhaps add parameter Qt::SkipEmptyParts?

usiems: Since we just had the case: What about adding a check if the given include directories actually exist? (This should include QTDIR!)

PR #168, opened 2023-12-14
More typesystem entries (added classes and rejections)

usiems:
Just so that there are no unknown rejected classes and enums in the log file for Qt 6.6. The support for some of the new classes might be incomplete.

One notable addition is the Cbor support in Qt. I got a little bit creative with the supporting classes like QCborTag (which is just a qint64 in Python now), and QCborNegativeInteger (which is simply not supported). Also the string reading methods of QCborStreamReader were modified, because PythonQt can't currently support arbitrary templated parameter/return types. If somebody wants to provide a patch to make it match the C++ interface...

I also re-allowed some methods in QtNetwork which might break compilation if Qt is compiled without SSL support. If this is a problem, please add a comment.

### DIFF
--- a/generator/typesystem_network.xml
+++ b/generator/typesystem_network.xml
@@ -5,10 +5,6 @@
   </namespace-type>
 
   <rejection class="QSqlError::Unused"/>
-  <rejection class="QSslConfiguration" function-name="defaultDtlsConfiguration"/>
-  <rejection class="QSslConfiguration" function-name="dtlsCookieVerificationEnabled"/>
-  <rejection class="QSslConfiguration" function-name="setDefaultDtlsConfiguration"/>
-  <rejection class="QSslConfiguration" function-name="setDtlsCookieVerificationEnabled"/>
 
   <enum-type name="QSsl::AlertLevel" since-version="6"/>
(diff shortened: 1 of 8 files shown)


generator/typesystem_network.xml
```diff
@@ -5,10 +5,6 @@
   </namespace-type>
 
   <rejection class="QSqlError::Unused"/>
-  <rejection class="QSslConfiguration" function-name="defaultDtlsConfiguration"/>
-  <rejection class="QSslConfiguration" function-name="dtlsCookieVerificationEnabled"/>
-  <rejection class="QSslConfiguration" function-name="setDefaultDtlsConfiguration"/>
-  <rejection class="QSslConfiguration" function-name="setDtlsCookieVerificationEnabled"/>
```
mrbean-bremen, line 11: I think we can safely assume that nowadays Qt is always compiled with SSL support. If that really should be a problem, we may just add a respective tag attribute to be able to exclude these.

mrbean-bremen (approved):
I didn't see anything in the rejections that looked like it would be needed, so this should be ok for now.

PR #176, opened 2023-12-15
Support top-level enum definitions

usiems:
Generate a wrapper for the global namespace if there is something to wrap (currently only enums are supported).

The enums in this wrapper are automatically copied to the Python package to which they belong, and the "Qt" package.

This was previously missing and made it hard to use e.g. QOcspResponse, because the required enum values were not generated.


mrbean-bremen (approved):
As far as I could understand this, it all looks good. Admittedly not easy to understand the logic of the generator here... Anyway, the proof of the pudding is in the eating, and it does work as intended.

PR #183, opened 2024-01-30
Add generated bindings for Qt 5.15

usiems:
Generated bindings for Qt 5.15.

It is somewhat hard to compare the generated code with the previous version, since we changed some general things (override instead of virtual), and because new classes push later classes to different header files. At a cursory inspection everything seems ok, though some generated methods may have been dropped (e.g. because they are inherited from a base class). Ah yes, and QRegularExpression is now a builtin type instead of QRegExp.

If you notice something is missing relative to Qt 5.11, please notify me.


mrbean-bremen (approved):

he-hesce:
Please tag an release candidate 2 on this master branch commit.

usiems:
Done.

he-hesce:
thanks.

PR #189, opened 2024-02-05
Force C++11 for older Qt versions with obsolete compilers (f.e. GCC 4.9.2)

iakov:
+ small fix for rare problem, but this reference management pattern is safer and preferable

### DIFF
--- a/src/PythonQt.cpp
+++ b/src/PythonQt.cpp
@@ -2137,8 +2137,9 @@ PyObject* PythonQtPrivate::packageByName(const char* name)
     v = PyImport_AddModule((_pythonQtModuleName + "." + name).constData());
     _packages.insert(name, v);
     // AddObject steals the reference, so increment it!
-    Py_INCREF(v);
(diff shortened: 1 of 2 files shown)


mrbean-bremen (approved):

iakov:
Any objections? Or can be merged? ;)

mrbean-bremen:
@usiems - can you have a look?

usiems:
Not sure if this might introduce problems down the line if Qt requires a greater C++ version than 11, but I guess we can fix that if problems arise.

usiems (approved):

iakov:
There is another problem with Qt6 that requires C++17. Even the generator project fails to compile with GCC 13 if `-ansi -pedantic -Werror` are in effect.  This is caused by Qt6 mkspecs.... Explicit `CONFIG+=c++14` does not influence command-line parameters. Only `CONFIG+=c++17` helps. As a compromise, the `CONFIG+=c++latest` can be used to be safe (as compatible with older compilers). But that is another story...

PR #199, opened 2024-03-29
Fix small issues in build and tests.

iakov:
* Main change: `CHAR_MIN` is `0` for unsigned `char`. In general, the "sign-ness" of the `char` type is implementation-specific. There is a proper check in the type conversion facility, but the test-case supposes the `char` to be signed, that is wrong.

* `opengl` is not required for tests (fix build without `opengl` module)
* Fix build with PYTHONQT_DEBUG enabled
* Small fixes to build with less warnings.

### DIFF
--- a/tests/PythonQtTests.cpp
+++ b/tests/PythonQtTests.cpp
@@ -193,7 +193,11 @@ void PythonQtTestSlotCalling::testPODSlotCalls()
   QVERIFY(_helper->runScript("if obj.getUInt(42)==42: obj.setPassed();\n"));
   QVERIFY(_helper->runScript("if obj.getShort(-43)==-43: obj.setPassed();\n"));
   QVERIFY(_helper->runScript("if obj.getUShort(43)==43: obj.setPassed();\n"));
+#if (CHAR_MIN + 0)
   QVERIFY(_helper->runScript("if obj.getChar(-12)==-12: obj.setPassed();\n"));
+#else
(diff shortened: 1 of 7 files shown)


tests/PythonQtTests.cpp
```diff
@@ -193,7 +193,11 @@ void PythonQtTestSlotCalling::testPODSlotCalls()
   QVERIFY(_helper->runScript("if obj.getUInt(42)==42: obj.setPassed();\n"));
   QVERIFY(_helper->runScript("if obj.getShort(-43)==-43: obj.setPassed();\n"));
   QVERIFY(_helper->runScript("if obj.getUShort(43)==43: obj.setPassed();\n"));
+#if (CHAR_MIN + 0)
```
usiems, line 196: If the "signedness" of char isn't fixed, I'm wondering if we should have a separate test for signed char then.
Does PythonQt even support this?

usiems (approved):

tests/PythonQtTests.cpp
iakov: I suggest (and this is implemented in the proposed patch) to test for `signed char` iff the implementation is signed, otherwise the test is performed for unsigned value grater than 128 (without loss of generality, I've used 250 in the patch).
If you suggest adding a check for signed char parameter overflow and unsigned char negative/overflow? Please, explain your idea to me, I'll add another check if it can be useful too.

PR #197, opened 2024-03-12
fix format-security

nim65s:
Hi,

This PR fix, on Arch, with GCC 13.2.1:

```cpp
make[1]: Entering directory '/home/nim/aur/pythonqt/src/pythonqt-3.5.0/generator'
g++ -c -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security -fstack-clash-protection -fcf-protection -Wp,-D_GLIBCXX_ASSERTIONS -Wno-deprecated-declarations -pedantic -Winit-self -Wuninitialized -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions -Wp,-D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security -fstack-clash-protection -fcf-protection -Wp,-D_GLIBCXX_ASSERTIONS -flto -fno-fat-lto-objects -Wall -Wextra -D_REENTRANT -fPIC -DRXX_ALLOCATOR_INIT_0 -DQT_NO_CAST_TO_ASCII -DQT_NO_DEBUG -DQT_GUI_LIB -DQT_XML_LIB -DQT_CORE_LIB -I. -I. -I. -I/home/nim/aur/pythonqt/src/pythonqt-3.5.0/generator/../common -Iparser -Iparser/rpp -Iparser/rpp -I/usr/include/qt -I/usr/include/qt/QtGui -I/usr/include/qt/QtXml -I/usr/include/qt/QtCore -I. -I/usr/lib/qt/mkspecs/linux-g++ -o main.o main.cpp
main.cpp:108:17: error: format not a string literal and no format arguments [-Werror=format-security]
  108 |         qWarning((reason + "This may cause problems with finding the necessary include files.").toUtf8().constData());
```

PR #207, opened 2024-04-16
fix format-security on Q_OS_MAC

nim65s:
Hi,

This is the following #197, except it is on macos with clang 16.0.6:
```
main.cpp:99:18: error: format string is not a string literal (potentially insecure) [-Werror,-Wformat-security]
        qWarning((reason + "Assuming standard binary install using frameworks.").toUtf8().constData());
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
main.cpp:99:18: note: treat the string as an argument to avoid this
        qWarning((reason + "Assuming standard binary install using frameworks.").toUtf8().constData());
                 ^
                 "%s",

### DIFF
--- a/generator/main.cpp
+++ b/generator/main.cpp
@@ -96,7 +96,7 @@ namespace
         QString reason = "The QTDIR environment variable " + qtdir.isEmpty() ?
                          "is not set. " : "points to a non-existing directory. ";
 #if defined(Q_OS_MAC)
-        qWarning((reason + "Assuming standard binary install using frameworks.").toUtf8().constData());
+        qWarning() << reason << "Assuming standard binary install using frameworks.";
             QString frameworkDir = "/Library/Frameworks";
             includes << (frameworkDir + "/QtXml.framework/Headers");
             includes << (frameworkDir + "/QtNetwork.framework/Headers");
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by mrbean-bremen — COMMENTED** [review]
Thanks! There is another `qWarning()` in main.cpp, and a couple in `typesystem.cpp`, can you adapt these too, while at it?

**Comment by nim65s** [comment]
Done, but as they didn't raise errors for me yet, I'm not sure if I can test those.

**Comment by nim65s** [comment]
At least, I was able to build on archlinux + nix (x86_64-linux + aarch64-darwin)

**Review by mrbean-bremen — APPROVED** [review]
Thanks! I think these are trivial enough to not need a test... maybe a second review instead.

**Comment by mrbean-bremen** [comment]
@usiems - please have another look!

File: generator/main.cpp [hunk 1]
```diff
@@ -80,7 +80,7 @@ namespace
       {
         if (!QDir(*it).exists())
         {
-          qWarning("Include path %s does not exist, ignoring it.", it->toUtf8().constData());
+          qWarning() << "Include path " << it->toUtf8() << "does not exist, ignoring it.";
```
usiems, line 83: I believe the second string should start with a space.

usiems (approved):

PR #221, opened 2024-06-02
Update install-qt-action to v4

iakov:
New environment variable `QT_ROOT_DIR` should be used instead of `Qt5_Dir`/`Qt6_Dir`.
Also, `PATH` is now correctly updated by install-qt-action, thus no workaround setting `GITHUB_PATH` is required. 
See https://github.com/jurplel/install-qt-action/blob/v4/README_upgrade_guide.md


mrbean-bremen (approved):

mrbean-bremen:
Thanks!

Issue #224, opened 2024-06-25
Qt 6.7: Several bindings are not generated


mrbean-bremen:
We noticed this only after making the patch release. 
Qt 6.7 binding are still quite incomplete, missing wrappers for QtCore include:
- QDeferredDeleteEvent
- QStandardPaths
- QStorageInfo
- QSystemSemaphore
- QTemporaryDir
- QTemporaryFile
- QTimeLine
- QTimer
- QTranslator
- QUrlQuery
- QWaitCondition
- QStreamEntityResolver

Also missing are operators for QDateTime and QTime.

usiems:
Should be fixed with merge request #225.

PR #225, opened 2024-06-27
Generator fixes for Qt 6.7 (#224)

usiems:
There was a bunch of problems, so looking at the commits separately is recommended.

Also not all commits solve real problems, I also tried to reduce parser errors somewhat.


generator/parser/binder.cpp
```diff
@@ -292,6 +292,19 @@ void Binder::declare_symbol(SimpleDeclarationAST *node, InitDeclaratorAST *init_
...
+      {
+        // check if this function declaration is a "friend" function.
+        // In this case we modify the scope, and remove the "friend" flag.
+        // This is for functions/operators that are only defined as friend
+        // inside of classes.
+        symbolScope = _M_current_file.staticCast<_ScopeModelItem>();
+        // unset the friend flag, as we treat this like a stand-alone function definition
+        fun->setFriend(false);
+        // also set the access policy to public, just in case
+        fun->setAccessPolicy(CodeModel::Public);
```
mrbean-bremen, line 305: Just in what case? Is this really needed?

generator/parser/parser.cpp
```diff
@@ -52,10 +52,22 @@
 #include <iostream>
 #include <vector>
 
+#define STRINGIZE(x) STRINGIZE2(x)
+#define STRINGIZE2(x) #x
+
+#if defined(_MSC_VER) || defined(__GNUC__)
+#define PARSER_METHOD_NAME __FUNCTION__ ":" STRINGIZE(__LINE__)
```
mrbean-bremen, line 59: Seems not to work for gcc...

generator/parser/binder.cpp
usiems: I simply copied this part from Binder::visitFunctionDefinition, where the same thing was done.

generator/parser/parser.cpp
usiems: Fixed.

mrbean-bremen (approved):

PR #228, opened 2024-08-09
Fix macOS and CentOS CI

MinyazevR:
1. [CentOS7 reached EOL](https://www.redhat.com/en/topics/linux/centos-linux-eol) and GHA support for the corresponding [container was abandoned](https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/). However, `node20` requires at least `glibc2.27`, which is not available in CentOS, but upgrading the glibc version contradicts the meaning of using older CentOS. The proposed solution in this PR is to use corresponding environment variable, that helps to extend CentOS CI availability for some time. Also,  `mirrorlist.centos.org` was turned off, so switch [to vault](https://serverfault.com/questions/904304/could-not-resolve-host-mirrorlist-centos-org-centos-7) is required (see `sed` for configs in PR)

2. [macOS-11 runner is no longer supported on GHA](https://github.blog/changelog/2024-05-20-actions-upcoming-changes-to-github-hosted-macos-runners/#macos-11-deprecation-and-removal), thus macOS-12 with python3.9 is proposed
 
3. Small feature: cancel running workflow from the same PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,7 +5,14 @@ on:
     branches: 
       - master
   pull_request:
-      
+   
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+   
+env:
+  ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
+    
 defaults:
   run:
     shell: bash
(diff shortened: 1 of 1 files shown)


mrbean-bremen:
@he-hesce - I think you are the one most interested in the CentOS builds. Can you have a look?

mrbean-bremen (approved):
Looks good to me, thanks! Just waiting for @he-hesce to comment.

he-hesce:
Looks good to me. RHEL 7 is still supported for several more years but we need the CentOS 7 container for testing on GitHub.

PR #229, opened 2024-08-09
Fix memory leak and add memory tests

MinyazevR:
1. [PyUnicode_FromString](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_FromString) returns a new reference, thus memory leak took place and is fixed.
2. To diagnose such leaks, basic memory tests are implemented, that to be extended later. To run these memory tests the  `PYTHONQT_RUN_ONLY_MEMORY_TESTS` environment variable should be used. 
3. These tiny specialized memory tests are temporary useful until memleaks are fixed in `PythonQt::cleanup` for general main tests.
4. So, this PR adds a CI step to run mentioned memory tests with Leak Sanitizer turned on via  `ASAN_OPTIONS=detect_leaks=1`.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -73,11 +73,13 @@ jobs:
       run: |
         QT_VERSION_FULL=$(qmake -query QT_VERSION)
         if [[ "$QT_VERSION_FULL" == 5.12* ]]; then
-          export LSAN_OPTIONS="suppressions=leak:QPlatformIntegrationFactory::create"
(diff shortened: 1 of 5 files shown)


iakov:
Thank you very much.
This PR is connected to #201 
Default tests leak a lot, that makes it difficult to test and fix leaks in the library itself. I agree that small separate test case is temporarily required, nevertheless the main test should be fixed too later.

usiems (approved):
Thank you for your effort, this all looks sensible.

Issue #230, opened 2024-08-13
QRegularExpression is missing since 3.5.0


tonka3000:
Hi,

I tried to use the newest tag `3.5.4`. Everything worked but `QRegularExpression` is missing from `PythonQt.QtCore`. It was the only class which is missing, all other classes where available. I use Qt 5.11.x in one of my project, so I only tested with this version of Qt.
The last working tag was `3.4.2`.

I was not able to track down the cause, but it seems that the restructuring (init for normal and builtin) from 3.5.x seems to be a reason because `QRegularExpression` is now in the builtin part. The `generated_cpp` has the calls for the registering, but not chance to import it at runtime.

usiems:
Did you rebuild PythonQt itself, with the newly generated builtin classes? Because they are part of the PythonQt library itself.

usiems:
I suspect that you built PythonQt with the old builtin code from the generated_cpp_511 folder, where QRegularExpressions didn't yet belong to the builtin classes (we didn't update that folder after all the generator changes). I remember vaguely that the build scripts automatically select this folder depending on the Python version. Try replacing the content of the generated_cpp_511 folder with the content from generated_cpp.

tonka3000:
So basically I should regenerate the bindings the generator, right?

It tried to remove the `generated_cpp_xxx` folders before building but it was screaming for the prebuilt ones.

I basically already generate the new bindings thanks to my conan recipe.
Not quiet sure how familiar you be with conan, here a snippet of the code.

```py
with tools.environment_append(env):
        self._run_qmake("PythonQt.pro")
        self._run_make()

        # generate qt bindings
        self.output.info("Generate Qt bindings")
        if self.settings.compiler == "Visual Studio":
            generator_path = os.path.join(self.qmake_config, self.generator_name)
        else:
            generator_path = "./%s" % self.generator_name

        with tools.chdir("generator"):
            self.run("%s qtscript_masterinclude.h build_all.txt " % (generator_path)) # build new bindings here
        
        # build qt bindings
        self.output.info("Build PythonQt extensions")
        with tools.chdir("extensions"):
            tools.replace_in_file("PythonQt_QtAll/PythonQt_QtAll.pro",
                "isEmpty",
                "PYTHONQTALL_CONFIG=$$(PYTHONQTALL_CONFIG)\nisEmpty"
            )
            tools.replace_in_file("PythonQt_QtAll/PythonQt_QtAll.pro",
[...]

PR #235, opened 2024-09-14
Add detailed build instructions to Readme

tonka3000:
Add more detailed build instructions to the Readme to avoid some common pitfalls (see #230) when building PythonQt.

For now I left the original build section in the Readme, but in my opinion the instructions can be removed.

Currently I mention of the env-vars I used for building it with conan, maybe some are not necessary in all configurations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,9 +38,10 @@ updating the typesystems as well.
 Building PythonQt requires a couple of steps.
 Follow these instructions in order to get a correctly built PythonQt runtime and Qt bindings.
 
-### ⚠️ Pregenerated Bindings Warning ⚠️
+### Recommendations
 
-It is highly recommended to build the Qt bindings yourself and not(!) use the pregenerated ones. To ensure that, delete the pregenerated directory starting with `generated_cpp_*`.
+It is recommended to build the Qt bindings yourself instead of using the pregenerated ones.
+This ensures the bindings are compatible with your Qt version.
 Do not build `PythonQt.pro` directly because it will only use the pregenerated bindings!
 
 ### Environment
@@ -115,15 +116,6 @@ As a last step, we need to build the extensions.
 
 After all these steps, you should now have a fully working PythonQt runtime and Qt bindings for your Python/Qt installation 🎉.
 
-## Building on Windows with MinGW
+## MinGW
 
-To build PythonQt, you need to set the environment variable `PYTHON_PATH` to
-point to the root dir of the python installation. Then you should set the
-`PYTHON_VERSION` variable to the Python version number.
-
-When using the prebuild Python installer, this will be:
-
-```cmd
-set PYTHON_PATH=c:\Python310
-set PYTHON_VERSION=3.10
-```
+It is possible to build PythonQt with MinGW on Windows instead of using MSVC.
--- a/README.md
+++ b/README.md
@@ -33,6 +33,88 @@ updating the typesystems as well.
 
 # Building
 
+## General
+
+Building PythonQt requires a couple of steps.
+Follow these instructions in order to get a correctly built PythonQt runtime and Qt bindings.
+
+### ⚠️ Pregenerated Bindings Warning ⚠️
+
+It is highly recommended to build the Qt bindings yourself and not(!) use the pregenerated ones. To ensure that, delete the pregenerated directory starting with `generated_cpp_*`.
+Do not build `PythonQt.pro` directly because it will only use the pregenerated bindings!
+
+### Environment
+
+First, you need to set a couple of environment variables, which depend on your Python and Qt installation.
+
+- `PYTHON_VERSION`
+
+  Set the version of your Python interpreter. The syntax is `<major>.<minor>`, e.g., `3.10`.
+
+- `PYTHON_PATH`
+
+  This is the absolute path to the root directory of your Python installation.
+
+- `PYTHON_LIB`
+
+  The absolute path to the `libs` directory in your python installation `$PYTHON_PATH/libs`.
+
+- `PYTHON_DIR`
+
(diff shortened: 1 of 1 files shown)


tonka3000:
@mrbean-bremen

README.md
```diff
@@ -33,6 +33,88 @@ updating the typesystems as well.
...
 
+## General
+
+Building PythonQt requires a couple of steps.
+Follow these instructions in order to get a correctly built PythonQt runtime and Qt bindings.
+
+### ⚠️ Pregenerated Bindings Warning ⚠️
+
+It is highly recommended to build the Qt bindings yourself and not(!) use the pregenerated ones. To ensure that, delete the pregenerated directory starting with `generated_cpp_*`.
+Do not build `PythonQt.pro` directly because it will only use the pregenerated bindings!
```
mrbean-bremen, line 44: Actually, it first checks for the path "generated_cpp", assuming that your self-generated sources are there, and only if it does not exist it checks for the other paths (see `common.prf`).

README.md
```diff
@@ -33,6 +33,88 @@ updating the typesystems as well.
...
 # Building
 
+## General
+
+Building PythonQt requires a couple of steps.
+Follow these instructions in order to get a correctly built PythonQt runtime and Qt bindings.
+
+### ⚠️ Pregenerated Bindings Warning ⚠️
+
+It is highly recommended to build the Qt bindings yourself and not(!) use the pregenerated ones. To ensure that, delete the pregenerated directory starting with `generated_cpp_*`.
```
mrbean-bremen, line 43: Maybe add some reasoning, like "to ensure that the bindings are always compatible with your Qt version".

README.md
```diff
@@ -33,6 +33,88 @@ updating the typesystems as well.
...
+
+   `qmake CONFIG+=Release extensions.pro`
+
+3. Make the extensions
+
+   Use `nmake` for MSVC (Visual Studio; make sure to have the environment variables set for Visual Studio beforehand). Otherwise, use `make`.
+
+After all these steps, you should now have a fully working PythonQt runtime and Qt bindings for your Python/Qt installation 🎉.
+
 ## Building on Windows with MinGW
```
mrbean-bremen, line 119: I think this part is now redundant. You may mention above that you can also build under Windows using `MinGW` and `make`.

mrbean-bremen (commented):
I have to say that I have not used qmake with PythonQt, just cmake, so I'm not the best to review this. Maybe @he-hesce can have another look...

README.md
he-hesce: From experience, 5.6 pre-gen bindings worked fine with 5.9, 5.12, and 5.15 albeit you never get the newer constructs and methods of course. 5.15 pre-gen bindings work fine with 5.15.

README.md
he-hesce: Indeed. No need to remove any pre-generated bindings. The generated_cpp will be used if created by user. I once tried to generate bindings and failed, so I've used 5.6 pre-gen bindings for a long time instead.

README.md
he-hesce: I have no experience with Windows so no idea if the above works.

he-hesce:
> I have to say that I have not used qmake with PythonQt, just cmake, so I'm not the best to review this. Maybe @he-hesce can have another look...

Well, I think those qmake lines will work though we build it differently. We place the pythonqt directory into our libsrc directory and add it to the libsrc.pro file in that directory. Then it gets built as part of the application build. It is then linked at the top-level project file. We do have to hack some of the lib paths in the PythonQt pro files for it to place build libraries where we place our other built libraries but that is extracurricular for our build environment/system. We do build our application including PythonQt using qmake which works fine and hopefully will continue to do so as we need to use and support qmake/qt5.15 for another decade or so.

README.md
mrbean-bremen: I still think it is a valid point - it is always preferrable to use the bindings generated for your concrete version instead of a pre-generated one for an older version. That generating the bindings did not work reliably before was one reason to use the pre-generated bindings that is now obsolete.
So maybe lessen the tone of the recommendation a bit ("It is recommended to build the Qt bindings yourself and not use the pregenerated ones."), and add a bit of a rational.

README.md
tonka3000: I used it with 5.11 and I got issues see #230

README.md
tonka3000: @mrbean-bremen Yes, it is redundant. I now mention that MinGW is an option on windows as you suggest.

README.md
tonka3000: @mrbean-bremen I added your mentioned reasion

README.md
tonka3000: I've lessen the tone.

tonka3000:
@he-hesce The qmake commands work because they are straight out of my conan recipe 😄 . I'm using 99% cmake, PyhonQt is the only project I use qmake for, so am not an expert in qmake. Also need to support 5.11 - 5.15 for now. But hopefully I make the jump to Qt 6 after having upgraded PythonQt to the lastest version.

README.md
mrbean-bremen: > Do not build `PythonQt.pro` directly because it will only use the pregenerated bindings!

This is partly true, in the sense that while it does build the generator first, it does not generate the bindings, so if you haven't built them before it will indeed use the checked-in ones. 
I think you can leave it as is, but ultimately PythonQt.pro shall be adapted to also generate the bindings, maybe depending on some variable (or add the CMakeLists.txt files, or both).

mrbean-bremen (approved):
I will wait with the merge a couple of days in case somebody else wants to chime in.

PR #236, opened 2024-10-02
Fix Qt 6.7.3

usiems:
This fixes the problem introduced with Qt 6.7.3, that the wrapper for the "Qt" namespace suddenly got methods like
` QKeyCombination  static_Qt_operator+(Qt::KeyboardModifiers  modifiers, Qt::Key  key);`
The problem is two-fold:

1. No proper name-mangling was applied to these static methods ("operator+" should be replaced by "\_\_add\_\_").
2. Static operators don't work in PythonQt anyway and shouldn't be generated (we also can't assign the operator to the first argument type, since that is only a enum type, which has no separate wrapper generated).

The trigger for these problems was that these static operators have been moved to the Qt namespace in version 6.7.3.

### DIFF
--- a/generator/shellgenerator.cpp
+++ b/generator/shellgenerator.cpp
@@ -316,7 +316,7 @@ bool ShellGenerator::functionHasNonConstReferences(const AbstractMetaFunction* f
       QTextStream t(&s);
       t << function->implementingClass()->qualifiedCppName() << "::";
       writeFunctionSignature(t, function, 0, "",
-        Option(ConvertReferenceToPtr | FirstArgIsWrappedObject | IncludeDefaultExpression | OriginalName | ShowStatic | UnderscoreSpaces | ProtectedEnumAsInts));
+        Option(ConvertReferenceToPtr | FirstArgIsWrappedObject | IncludeDefaultExpression | ShowStatic | UnderscoreSpaces | ProtectedEnumAsInts));
       std::cout << s.toLatin1().constData() << std::endl;
       return true;
     }
--- a/generator/typesystem_core.xml
+++ b/generator/typesystem_core.xml
@@ -1507,9 +1507,7 @@ public:
   <object-type name="QSemaphore"/>
   <object-type name="QSocketNotifier"/>
   <object-type name="QSystemLocale"/>
-  <object-type name="QTemporaryFile">
-    <modify-function signature="fileName()const" rename="uniqueFilename"/>
-  </object-type>
+  <object-type name="QTemporaryFile"/>
   <object-type name="QMimeData">
     <extra-includes>
       <include file-name="QStringList" location="global"/>
--- a/generator/typesystem_gui.xml
+++ b/generator/typesystem_gui.xml
@@ -797,9 +797,6 @@
   <value-type name="QTextFormat">
     <modify-function signature="operator=(QTextFormat)" remove="all"/>
     <modify-function signature="isValid()const" access="non-final"/>
-
-    <modify-function signature="setProperty(int,QVector&lt;QTextLength&gt;)" rename="setLengthVectorProperty" before-version="6"/>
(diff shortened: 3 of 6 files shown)


generator/typesystem_gui.xml
```diff
@@ -917,10 +914,6 @@
 
       <modify-function signature="operator*(QMatrix)const" access="private"/>
       <modify-function signature="operator*=(QMatrix)" access="private"/>
-      <modify-function signature="rotate(double)" access="private" rename="rotate_private"/>
-      <modify-function signature="scale(double,double)" access="private" rename="scale_private"/>
-      <modify-function signature="shear(double,double)" access="private" rename="shear_private"/>
-      <modify-function signature="translate(double,double)" access="private" rename="translate_private"/>
```
mrbean-bremen, line 923: What have these been for? Is this pre-Qt5 stuff?

generator/typesystem_gui.xml
```diff
@@ -1562,8 +1555,6 @@ PyObject* constScanLine(QImage* image, int line) {
   <object-type name="QTextBlockUserData"/>
   <object-type name="QTextItem"/>
   <object-type name="QTextList">
-    <modify-function signature="format()const" rename="textListFormat"/>
```
mrbean-bremen, line 1565: What about compatibility? (same below)

generator/shellgenerator.cpp
```diff
@@ -316,7 +316,7 @@ bool ShellGenerator::functionHasNonConstReferences(const AbstractMetaFunction* f
       QTextStream t(&s);
       t << function->implementingClass()->qualifiedCppName() << "::";
       writeFunctionSignature(t, function, 0, "",
-        Option(ConvertReferenceToPtr | FirstArgIsWrappedObject | IncludeDefaultExpression | OriginalName | ShowStatic | UnderscoreSpaces | ProtectedEnumAsInts));
+        Option(ConvertReferenceToPtr | FirstArgIsWrappedObject | IncludeDefaultExpression | ShowStatic | UnderscoreSpaces | ProtectedEnumAsInts));
```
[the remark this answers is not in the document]
usiems: Basically the option said: "Don't use the name that was explicitly set for this method, but use the name from the parsed C++ code." This made a difference for operators, which were set to the name of the Python operators in the AbstractMetaBuilder step. There was code which did this replacement again in the code generator, but this didn't apply to static operators.
I didn't see the purpose of this option, since it also reverted some other renames supplied by the typesystem files, so I removed it. Now the code should behave more straightforward.

generator/typesystem_core.xml
```diff
@@ -1507,9 +1507,7 @@ public:
   <object-type name="QSemaphore"/>
   <object-type name="QSocketNotifier"/>
   <object-type name="QSystemLocale"/>
-  <object-type name="QTemporaryFile">
-    <modify-function signature="fileName()const" rename="uniqueFilename"/>
```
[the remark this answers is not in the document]
usiems: No, these renames were ignored because of the OriginalName option.
One reason more why I removed the OriginalName option.

generator/typesystem_gui.xml
usiems: I don't know, at least since Qt5 these variants don't even exist (the arguments are all float now according to the documentation). And PythonQt should have no problems to discern between float and double variants of the same method.

generator/typesystem_gui.xml
usiems: See above, the rename wasn't applied.

generator/shellgenerator.cpp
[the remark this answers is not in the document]
usiems: (I did a before/after comparison of the generated code, and the only difference was that some wrapper methods were not created anymore that are also provided by the super class, and this was because I removed the previously ignored rename entries.)

mrbean-bremen (approved):

PR #242, opened 2024-12-02
opengl compatible Qt5 and Qt6

wangdaye078:


extensions/PythonQt_QtAll/PythonQt_QtAll.pro
```diff
@@ -115,6 +112,9 @@ PythonQtNetwork {
 PythonQtOpengl {
   DEFINES += PYTHONQT_WITH_OPENGL
   QT += opengl
+  equals(QT_MAJOR_VERSION, 6){
```
mrbean-bremen, line 115: maybe better use `greaterThan(QT_MAJOR_VERSION, 5)` here

mrbean-bremen (approved):
As far as I can see this looks good, thanks! Though I would like that @usiems has another look.
You can ignore the failing tests for the OldSchool workflows, they are unrelated.

mrbean-bremen:
Ping @usiems

usiems:
Sorry for my late answer.
Of course these changes are correct in principle. What I don't like about them is that they would break backward compatibility to the previous PythonQt version, as some classes would now appear in PythonQt.QtOpenGL instead of PythonQt.QtGui.
Ideally these classes would appear in both packages: The new place, where they would be expected according to the Qt 6 documentation, and - perhaps optionally - the old place, so existing code doesn't need to be changed. I don't believe PythonQt currently supports this, and I don't know how much effort it would be to implement this.

PR #247, opened 2024-12-30
Remove no longer supported CentOS 7 builds

mrbean-bremen:
- see https://github.com/actions/runner/issues/2906

Node16 support has been completely removed from GH actions, together with CentOS 7 support, and the workaround no longer works.
There are more hacky workarounds to get it running again, but I'm reluctant to use them. Any installation on these systems should be fine using a released version - further changes in PythonQt will be most likely be related to Qt 6 only.
CC @he-hesce.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,9 +10,6 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
    
-env:
-  ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
-    
 defaults:
   run:
     shell: bash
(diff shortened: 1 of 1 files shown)


he-hesce:
Ok if you skip CentOS 7 testing as CentOS 7 is out of support. RHEL 7 is still in support and in use though for several more years. We are still using PythonQt on RHEL 7 so please avoid making any breaking changes such as killing Qt 5 support or qmake support. It is practical to be able to use the same PythonQt version on all platforms and not having to have separate branching. Happy New Year PythonQt community!

mrbean-bremen:
Sure - no sense in breaking compatibility if there is no real reason to that. This is only about the CI. Feel free to add more CI checks if you need them :)
Happy New Year to you too!

PR #270, opened 2025-07-16
BUG: Remove deprecated Qt6 paradigms

hjmjohnson:
Replace with backward compatible recommended or
preprocessor conditional for old and new.

Use new paradigms that are backward compatible
when possible.

Use preprocessor conditionals where the syntax
is different between qt5 and qt6.


jcfr:
This pull request should be closed. It is superseded by:
* https://github.com/MeVisLab/pythonqt/pull/278

PR #272, opened 2025-07-16
COMP: Remove use of deprecated 'count()' function

hjmjohnson:
PythonQt/generator/parser/name_compiler.cpp:111:46: \
    warning: 'count' is deprecated: Use size() or length() instead.
    [-Wdeprecated-declarations]


mrbean-bremen (approved):
Thanks!

Issue #276, opened 2025-08-28
Qt6 on macos: missing generated_cpp


nim65s:
Hi again,

I am trying to build pythonqt for Qt6 on linux and macos to [update the package on nixpkgs](https://github.com/NixOS/nixpkgs/pull/437521), but I am not able to generate the cpp on macos.

This issue can already be seen in the current CI: 
https://github.com/MeVisLab/pythonqt/actions/runs/16336586053/job/46149827921 :
```
Run cd generator
Please wait while source files are being generated...
Parsing typesystem file [:/trolltech/generator/build_all.txt]
"is not set. " Assuming standard binary install using frameworks.
WARNING(Generate) :: ShellImplGenerator: no java classes, skipping
WARNING(Generate) :: ShellHeaderGenerator: no java classes, skipping
PreProcessing - Generate [.preprocessed.tmp] using [:/trolltech/generator/qtscript_masterinclude.h] and include-paths [/Users/runner/work/pythonqt/Qt/5.9.9/clang_64/lib]
Building model using [.preprocessed.tmp]
Classes in typesystem: 0
Generated:
  - header....: 0 (0)
  - impl......: 0 (0)
  - modules...: 0 (0)
  - pri.......: 0 (0)

Done, 2 warnings (1851 known issues)
0s
Run actions/upload-artifact@v4
Warning: No files were found with the provided path: generated_cpp. No artifacts will be uploaded.
```

But this does not trigger any failure, so I guess this was not seen.

I tried to add a more explicit CI failure in #275 and #277, but that failed.

Anyways, could this be done ? In that case, I guess we should add macos to the "check generated_cpp" CI. Otherwise, we should probably add the generated_cpp for qt 6 to git.

PR #277, opened 2025-08-28
CI: check generated_cpp on macos

nim65s:
Trying again for #276


nim65s:
Ok, this is not supported by https://github.com/jurplel/install-qt-action#arch

PR #278, opened 2025-08-28
chore: Remove deprecated Qt5 paradigms

jcfr:
This changes deprecated Qt6 paradigms to either backward-compatible ones or uses preprocessor conditionals for Qt5 and Qt6 differences. This ensures compatibility with both Qt versions.

---

This pull request supersedes https://github.com/MeVisLab/pythonqt/pull/270


hjmjohnson (commented):
Thanks @jcfr

PR #282, opened 2025-08-28
Let the CI fail if no wrappers are generated

mrbean-bremen:
- see #276


mrbean-bremen:
As discussed: commented out for the failing tasks, shall be re-added after #276 and #291 are fixed.

PR #285, opened 2025-08-30
Re-introduce support for QMatrix when building against Qt < 6.0

jcfr:
This partially reverts 10a3e499 ("Qt6 compatibility fixes for core PythonQt library", 2023-08-21)

(cherry picked from commit commontk/PythonQt@b2156e9b1b61d35fb3ef19fd5aadeb2246cf9ff8)


mrbean-bremen (approved):
Thanks - this was obviously missed in the transition to Qt6.

PR #286, opened 2025-08-30
Improve `stdin`, `stdout`, and `stderr` redirection in PythonQt

jcfr:
This pull request introduces and fixes the handling of standard stream redirection in the PythonQt module.

- Fix incorrect conditions in `setRedirectStdInCallbackEnabled` to ensure proper `stdin` redirection.
- Add support for redirecting `stdout` and `stderr` to custom callbacks.
- Introduce methods to enable or disable redirection and to query the current redirection state.

> [!NOTE]
> For reference, those patches were developed in the context of the `commontk/PythonQt` fork.

### DIFF
--- a/src/PythonQt.cpp
+++ b/src/PythonQt.cpp
@@ -502,11 +502,11 @@ void PythonQt::setRedirectStdInCallbackEnabled(bool enabled)
   sys.setNewRef(PyImport_ImportModule("sys"));
 
   if (enabled) {
-    if( !PyObject_HasAttrString(sys.object(), "pythonqt_stdin") ) {
+    if( PyObject_HasAttrString(sys.object(), "pythonqt_stdin") ) {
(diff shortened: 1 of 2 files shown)


src/PythonQt.cpp
```diff
@@ -502,11 +502,11 @@ void PythonQt::setRedirectStdInCallbackEnabled(bool enabled)
   sys.setNewRef(PyImport_ImportModule("sys"));
 
   if (enabled) {
-    if( !PyObject_HasAttrString(sys.object(), "pythonqt_stdin") ) {
+    if( PyObject_HasAttrString(sys.object(), "pythonqt_stdin") ) {
```
mrbean-bremen, line 505: I don't understand how this has worked before - also would ask @usiems to review, who was been involved here.

usiems: The commit I reverted was about memory leaks (which caused crashes, thus the revert). I didn't realize that the commit also included another bug fix.

PR #299, opened 2025-09-12
Fix warning about missing or non-existing QTDIR

mrbean-bremen:
- missing parenthesis caused wrong warning output

The warning was always "is not set. " instead of "The QTDIR environment variable is not set. " or "The QTDIR environment variable points to a non-existing directory. "


mrbean-bremen:
Superseded by #302.

PR #306, opened 2025-09-20
chore: Remove obsolete Python 2 compatibility code

jcfr:
This pull request removes obsolete compatibility code for Python 2 and outdated 
preprocessor conditionals for Python versions < 2.4.

* Remove all `#ifndef PY3K` conditionals and related branches, consolidating on Python 3 APIs. This drops redundant code paths and consistently renames `nb_nonzero` helpers to `nb_bool`.

* Remove the comment related to the use of `Py_FlushLine` in `custom_system_exit_exception_handle` in `PythonQt.cpp`, which is not needed in Python 3.x. See python/cpython@79139b247b0 ("Kill off softspace completely (except in formatter.py which seems to have a different feature with the same name). The change to test_doctest.txt reduces the doctest failures to 3.", 2007-02-09)

* Remove outdated `PY_VERSION_HEX` conditional branches.

### DIFF
--- a/src/PythonQt.cpp
+++ b/src/PythonQt.cpp
@@ -101,7 +101,6 @@ void PythonQt::init(int flags, const QByteArray& pythonQtModuleName)
       _self->_p->_pySourcelessFileLoader = importlib.getVariable("SourcelessFileLoader");
     }
 
-#ifdef PY3K
     // Import asyncio only when not explicitly disabled.
     // Importing asyncio on Py3.12+ pulls in ssl/_ssl; some environments/tests
     // want to avoid that during early embedded init.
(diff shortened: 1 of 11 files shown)


mrbean-bremen (approved):
Thanks - it's time to really declare Python 2 dead...

jcfr:
Ditto. Have a few more related PRs coming up 🚀

he-hesce:
You should probably mark a final version as compatible with Python 2.7 so those who needs Python 2 support knows what to use - and then do a significant version bump to signal Python 3 only support.

Note that Python 2.7 is still the default version and supported on RHEL 7.9 which is in Extended Life Cycle Support phase until 30 June 2028.

mrbean-bremen:
I think all the latest changes have been for Qt 5.15 and 6, which is not what is used on these systems. So basically, all changes since version 3.5 are basically targeted to Qt 5.15 and later, and Python 3.
The documentation here is a bit sparse, to say the least, and there are no separate release notes, so the only place to put this kind of information are the GitHub releases, and maybe the README. 

That being said, I agree that a major version update makes sense after the synchronization with the commontk version is finished, with a note that Python 2 is no longer supported, and versions before Qt 5.15 are no longer guaranteed to work.

PR #308, opened 2025-09-25
chore(PythonQt): Migrate to Python-3 Unicode API; update generator/wrappers; add opt-in PyString shim; enforce Python ≥ 3

jcfr:
This PR completes the migration to Python 3 string APIs across PythonQt, updates the code generator and generated wrappers, and introduces an **opt-in** compatibility shim for legacy wrappers.

It removes reliance on `PyString_*` aliases in `PythonQtPythonInclude.h` for in-tree sources and makes Unicode usage explicit. The build now **errors** if Python 2.x is detected.

**Summary:**

* Replace `PyString_*` with Python-3 equivalents:

  * `PyString_FromString` → `PyUnicode_FromString`
  * `PyString_AS_STRING` → `PyUnicode_AsUTF8`
  * `PyString_AsString` → `PyUnicode_AsUTF8`
  * `PyString_FromFormat` → `PyUnicode_FromFormat`
  * `PyString_Check` → `PyUnicode_Check`
* Update the **generator** to emit `PyUnicode_*`.
* Update **generated_cpp_*** wrappers to use `PyUnicode_*`.
* Add `PYTHONQT_USE_PYSTRING_SHIM` (default **OFF**) to alias `PyString_FromString` → `PyUnicode_FromString` for older, out-of-tree wrappers.
* Enforce Python 3 at build time (hard error on Python 2.x).


jcfr:
> Build / oldschool (rockylinux, 9, debug) (pull_request)

Reported error[^1] is unrelated to the proposed changes:

[^1]: https://github.com/MeVisLab/pythonqt/actions/runs/18019484218/job/51273010918?pr=308

```
Complete!
Last metadata expiration check: 0:00:55 ago on Thu Sep 25 20:13:12 2025.
Package which-2.21-30.el9_6.x86_64 is already installed.
Error: 
 Problem: package qt5-qtbase-private-devel-5.15.9-11.el9_6.x86_64 from appstream requires cups-devel, but none of the providers can be installed
  - cannot install the best candidate for the job
  - nothing provides cups-libs(x86-32) = 1:2.3.3op2-33.el9_6.1 needed by cups-devel-1:2.3.3op2-33.el9_6.1.i686 from appstream
  - nothing provides cups-libs(x86-64) = 1:2.3.3op2-33.el9_6.1 needed by cups-devel-1:2.3.3op2-33.el9_6.1.x86_64 from appstream
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
```

generated_cpp_50/com_trolltech_qt_core/com_trolltech_qt_core0.cpp
```diff
@@ -50,7 +50,7 @@ PythonQtShell_QAbstractAnimation::~PythonQtShell_QAbstractAnimation() {
 void PythonQtShell_QAbstractAnimation::childEvent(QChildEvent*  arg__1)
```
mrbean-bremen, line 50: I wonder if we should remove these older versions of generated code now, instead of adapting them to these changes each time. 
Qt versions before 5.9 are not tested at all, and we do not test the checked-in generated code anyway, so we cannot guarantee that it will still work with the current version of PythonQt. Had also a discussion with @usiems about this.

src/PythonQtPythonInclude.h
```diff
@@ -117,31 +117,23 @@
...
+#if PY_MAJOR_VERSION < 3
+#error "PythonQt requires Python >= 3.x"
+#endif
+
+// Optional compatibility shim for legacy wrappers generated by older PythonQt.
+// Enable by defining PYTHONQT_USE_PYSTRING_SHIM (deprecated).
+#if defined(PYTHONQT_USE_PYSTRING_SHIM)
 #define PY3K
-// Helper defines to facilitate porting
 #define PyString_FromString PyUnicode_FromString
```
mrbean-bremen, line 128: Maybe add a deprecation warning here? Or would this over-flood the log with warnings?

PR #309, opened 2025-09-25
chore: Replace deprecated Qt macros and Improve const correctness

jcfr:
This pull request includes two main changes:
1. Replaces deprecated `foreach` and `Q_FOREACH` macros with C++11 range-based for loops as recommended by Qt 5.7+.
2. Improves const correctness by using `const` references in range-based for loops and other relevant places.

### DIFF
--- a/generator/abstractmetalang.cpp
+++ b/generator/abstractmetalang.cpp
@@ -158,7 +158,7 @@ AbstractMetaFunction::~AbstractMetaFunction()
 bool AbstractMetaFunction::isModifiedRemoved(int types) const
 {
     FunctionModificationList mods = modifications(implementingClass());
-    foreach (FunctionModification mod, mods) {
+    for (FunctionModification mod : mods) {
         if (!mod.isRemoveModifier())
             continue;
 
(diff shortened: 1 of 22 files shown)


generator/parser/binder.cpp
```diff
@@ -331,7 +331,8 @@ void Binder::declare_symbol(SimpleDeclarationAST *node, InitDeclaratorAST *init_
       fun->setVariadics (decl_cc.isVariadics ());
 
       // ... and the signature
-      foreach (DeclaratorCompiler::Parameter p, decl_cc.parameters())
+      const QList<DeclaratorCompiler::Parameter> parameters = decl_cc.parameters();
```
mrbean-bremen, line 334: Would `const auto` work here? If we are using C++11 as we do, we could really use `auto` a bit more (though there are certainly cases which only work from C++17 onwards).

generator/parser/binder.cpp
```diff
@@ -331,7 +331,8 @@ void Binder::declare_symbol(SimpleDeclarationAST *node, InitDeclaratorAST *init_
...
+      for (const DeclaratorCompiler::Parameter& p : parameters)
         {
```
mrbean-bremen, line 336: Same here (and in the following changes).

generator/abstractmetalang.cpp
```diff
@@ -811,15 +812,15 @@ AbstractMetaClass *AbstractMetaClass::extractInterface()
 
         iface->setTypeEntry(typeEntry()->designatedInterface());
 
-        foreach (AbstractMetaFunction *function, functions()) {
+        for (const AbstractMetaFunction* function : this->functions()) {
```
mrbean-bremen, line 815: No reason no use the explicit `this` here and below

usiems (approved):

PR #311, opened 2025-09-25
chore(importer): read `sys.flags` (verbose/optimize) on Python >= 3.12; fallback to legacy globals

jcfr:
Python 3.12 deprecates direct use of global flags (`Py_VerboseFlag`, `Py_OptimizeFlag`). Query `sys.flags.verbose` / `sys.flags.optimize` at runtime (via a small helper) when building against 3.12+, and fallback to the legacy globals on older versions.

No functional change intended.

### DIFF
--- a/src/PythonQtImporter.cpp
+++ b/src/PythonQtImporter.cpp
@@ -80,6 +80,26 @@ struct st_mlab_searchorder {
 extern PyTypeObject PythonQtImporter_Type;
 PyObject *PythonQtImportError;
 
+namespace
+{
+int getSysFlag(const char* flag_name)
+{
+  PyObject* flags = PySys_GetObject("flags");
+  if (!flags) {
+    return false;
+  }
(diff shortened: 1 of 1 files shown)


src/PythonQtImporter.cpp
```diff
@@ -308,7 +328,11 @@ PythonQtImporter_load_module(PyObject *obj, PyObject *args)
     }
 
     Py_DECREF(code);
+#if PY_VERSION_HEX >= 0x030C0000  // Python >= 3.12
+    if (getSysFlag("verbose")) {
```
mrbean-bremen, line 332: Probably makes sense to add convenience functions for this often used flag (and maybe other flags), instead of cluttering the code with `#ifdef`s.

mrbean-bremen (approved):

usiems (approved):

PR #319, opened 2025-10-26
fix: Ensure typesystem settings can be loaded from Qt resources and update CI workflow

jcfr:
This pull request improves the way typesystem settings are loaded by ensuring they can be referenced as Qt resources. It updates the `build_all.txt` to use Qt resource paths and adds the missing `typesystem_uitools.xml` entry.

Additionally, the CI workflow is updated to run the generator outside of the source directory, ensuring the changes are effective. The output directory is explicitly specified to be the current directory to avoid issues with default output paths.


mrbean-bremen (approved):
Looks like an oversight in the previous implementation - the resources have already been there...

PR #326, opened 2025-11-14
Add clang-format configuration

MinyazevR:
Full description: https://github.com/MeVisLab/pythonqt/pull/265

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -7,7 +7,7 @@ AccessModifierOffset: -2
 ReflowComments:  false
 ConstructorInitializerIndentWidth: 2
 ContinuationIndentWidth: 2
-AllowShortFunctionsOnASingleLine: None
+AllowShortFunctionsOnASingleLine: Inline
 BreakBeforeBraces: Custom
 BraceWrapping:
     AfterExternBlock: true
@@ -15,8 +15,8 @@ BraceWrapping:
     AfterFunction: true
     AfterNamespace: false
     AfterEnum: false
-    AfterCaseLabel: false  
-    AfterControlStatement: Never
+    AfterCaseLabel: true
+    AfterControlStatement: MultiLine
     AfterStruct:     false
     AfterUnion:      false
     BeforeLambdaBody: false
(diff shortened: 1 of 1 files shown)


.clang-format
```diff
@@ -0,0 +1,174 @@
...
+    BeforeElse: false
+    BeforeCatch:     false
+    BeforeWhile:     false
+    IndentBraces:    false
+    SplitEmptyFunction: true
+    SplitEmptyRecord: false
+    SplitEmptyNamespace: true
+PPIndentWidth: 2
+FixNamespaceComments: false
+# ??????
```
mrbean-bremen: ?

mrbean-bremen (commented):
Thanks, this looks good! I would like a couple of minor changes, and would like @usiems to have another look, before merging this.

.clang-format
```diff
@@ -0,0 +1,173 @@
...
+ContinuationIndentWidth: 2
+AllowShortFunctionsOnASingleLine: None
+BreakBeforeBraces: Custom
+BraceWrapping:
+    AfterExternBlock: true
+    AfterClass: true
+    AfterFunction: true
+    AfterNamespace: false
+    AfterEnum: false
+    AfterCaseLabel: false
```
usiems: 'AfterCaseLabel: true' better matches the current style.

I rather dislike that the `break` behind a closing brace is put onto the same line, but found no option to influence this. The developers of clang-format seem to believe that this is as it should be.

run-clang-format.sh
jamesobutler: You may want to consider adding the clang-format hook to a pre-commit-config yaml as 3D Slicer and other related projects have adopted (see https://github.com/Slicer/Slicer/blob/1b1bbb22b4f858ff3d47d196de4ac491a0b9ea5c/.pre-commit-config.yaml#L27-L39). Pre-commit-config yml usage has been adopted by many projects and something that is generally called in a GitHub actions workflow file (e.g. lint.yml) using the pre-commit github action (https://github.com/pre-commit/action).

^This way new styling is applied consistently and enforced for every PR.

usiems (approved):

PR #345, opened 2026-03-29
Add cmake build system

jamesobutler:
This PR is an effort previously discussed to bring in the CMake build system for modern cross compilation of C++ libraries. This PR includes the CMake build system as of https://github.com/commontk/PythonQt/tree/patched-v4.0.1-2026-03-16-5cd9b581f. The CommonTK organization is used by applications such as [3D Slicer](https://github.com/Slicer/Slicer) and [MITK](https://github.com/MITK/MITK). Using CMake to build PythonQt has been utilized successfully by 3D Slicer for well over a decade. Integration of this work back to the upstream has taken a long time due to multiple factors including MeVisLab's PythonQt source repository not on GitHub and 3D Slicer maintaining a series of patches that had not been contributed back to the upstream. Both situations have changed where PythonQt is on GitHub and the long list of patches were contributed to the upstream in the fall of 2025.

A large majority of this CMake work is due to the fantastic work by @jcfr who also worked to contribute the patches back to this upstream repo. Late last year JC made a job move to Nvidia and therefore contributions to 3D Slicer and the ecosystem of supporting open-source libraries has been limited. I submit this PR to keep this work going to bring the CommonTK fork of PythonQt back in sync with the MeVisLab PythonQt upstream. 

I provide all the commits here for the CMake development for historic reference, but maintainers could consider squashing of commits for less noise. 

I have created a new GitHub actions workflow as part of testing the CMake build system similar to the existing CI infrastructure. This PR doesn't currently remove the existing build system, though maintainers here could consider a replacement based on stabilization here and then avoid the complexity of maintaining 2 build systems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,275 @@
+
+cmake_minimum_required(VERSION 3.20.6)
+
+project(PythonQt)
+
+#----------------------------------------------------------------------------
+# Qt version
+
+set(_default_qt_major_version 5)
+if(DEFINED Qt6_DIR)
+  set(_default_qt_major_version 6)
+endif()
+
+# Set PythonQt_QT_VERSION
+set(PythonQt_QT_VERSION ${_default_qt_major_version} CACHE STRING "Qt major version (5 or 6)")
+set_property(CACHE PythonQt_QT_VERSION PROPERTY STRINGS "5" "6")
+if(NOT "${PythonQt_QT_VERSION}" MATCHES "^(5|6)$")
+  message(FATAL_ERROR "error: PythonQt_QT_VERSION must be 5 or 6.")
+endif()
+
+# Requirements
+set(minimum_required_qt5_version "5.15.0")
+set(minimum_required_qt6_version "6.4.0")
+set(minimum_required_qt_version ${minimum_required_qt${PythonQt_QT_VERSION}_version})
+
+find_package(Qt${PythonQt_QT_VERSION} ${minimum_required_qt_version} QUIET)
+
+#----------------------------------------------------------------------------
+# Qt components
+set(qtlibs
+  Core
+  Widgets
+  Network
+  OpenGL
+  Qml
+  Quick
+  Sql
+  Svg
+  Multimedia
+  WebEngineWidgets
+  UiTools
+  Xml
+  )
+# XmlPatterns and WebKitWidgets are removed from Qt 6
+if(PythonQt_QT_VERSION VERSION_EQUAL "5")
+  list(APPEND qtlibs
+    XmlPatterns
+    WebKitWidgets
+  )
+endif()
+
+#-----------------------------------------------------------------------------
+# Python libraries
+
+find_package(Python3 COMPONENTS Development REQUIRED)
+
+#-----------------------------------------------------------------------------
+# Build options
+
+option(PythonQt_SUPPORT_NAME_PROPERTY "Enable PythonQt name-property support" ON)
+option(PythonQt_USE_RELEASE_PYTHON_FALLBACK "Fallback to Release python when Debug missing" ON)
+option(PythonQt_DEBUG "Enable PythonQt debug output" OFF)
+option(PythonQt_BUILD_QTALL "Build PythonQt_QtAll bindings as a separate library" ON)
+
+if(NOT DEFINED PythonQt_INSTALL_RUNTIME_DIR)
+  set(PythonQt_INSTALL_RUNTIME_DIR bin)
+endif()
+
+if(NOT DEFINED PythonQt_INSTALL_LIBRARY_DIR)
+  set(PythonQt_INSTALL_LIBRARY_DIR lib${LIB_SUFFIX})
+endif()
+
+if(NOT DEFINED PythonQt_INSTALL_ARCHIVE_DIR)
+  set(PythonQt_INSTALL_ARCHIVE_DIR lib${LIB_SUFFIX})
+endif()
+
+if(NOT DEFINED PythonQt_INSTALL_INCLUDE_DIR)
+  set(PythonQt_INSTALL_INCLUDE_DIR include/PythonQt)
+endif()
+
+#-----------------------------------------------------------------------------
+# Set qtlib_to_wraplib_* variables
+
+set(qtlib_to_wraplib_Widgets gui)
+set(qtlib_to_wraplib_WebKitWidgets webkit)
+
+set(qt_wrapped_lib_depends_gui Multimedia PrintSupport)
+if(PythonQt_QT_VERSION VERSION_EQUAL "6")
+  list(APPEND qt_wrapped_lib_depends_gui Widgets OpenGL)
+endif()
+set(qt_wrapped_lib_depends_multimedia MultimediaWidgets)
+set(qt_wrapped_lib_depends_quick QuickWidgets)
+set(qt_wrapped_lib_depends_svg )
+if(PythonQt_QT_VERSION VERSION_EQUAL "6")
+  list(APPEND qt_wrapped_lib_depends_svg SvgWidgets)
+endif()
+set(qt_wrapped_lib_depends_webenginewidgets )
+set(qt_wrapped_lib_depends_webkit )
+
+foreach(qtlib ${qtlibs})
+  string(TOLOWER ${qtlib} qtlib_lowercase)
+  if(DEFINED qtlib_to_wraplib_${qtlib})
+    set(qtlib_lowercase ${qtlib_to_wraplib_${qtlib}})
+  endif()
+  set(qtlib_to_wraplib_${qtlib} ${qtlib_lowercase})
+endforeach()
+
+#-----------------------------------------------------------------------------
+# Define PythonQt_Wrap_Qt* options
+option(PythonQt_Wrap_QtAll "Make all Qt components available in python" OFF)
+foreach(qtlib ${qtlibs})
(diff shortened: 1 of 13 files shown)


src/PythonQtClassInfo.cpp
usiems, line 1: Can we please have a separate merge request for the flags stuff? Since this is totally unrelated to the CMake stuff?

src/generate_configure_h.py
```diff
@@ -0,0 +1,37 @@
...
+        )
+        sys.exit(1)
+
+    in_path, out_path = sys.argv[1], sys.argv[2]
+
+    with open(in_path) as f:
+        content = f.read()
+
+    # #cmakedefine VAR [rest]  ->  #define VAR [rest]
+    content = re.sub(r"^#cmakedefine\b", "#define", content, flags=re.MULTILINE)
```
usiems: It seems to me as if this script always changes #cmakedefine into #define, so there is no way to opt out of this option when using qmake. Too retain backwards compatibility it should be changed into something which does _not_ set this variable.

CMakeLists.txt
```diff
@@ -0,0 +1,420 @@
...
+  Network
+  OpenGL
+  Qml
+  Quick
+  Sql
+  Svg
+  Multimedia
+  UiTools
+  Xml
+  )
```
usiems, line 43: It seems to me that WebEngineWidgets is missing from this list.

CMakeLists.txt
```diff
@@ -0,0 +1,420 @@
...
+    )
+
+  target_compile_definitions(PythonQtCppTests
+    PRIVATE
+      $<$<BOOL:${PythonQt_SUPPORT_NAME_PROPERTY}>:PYTHONQT_SUPPORT_NAME_PROPERTY>
+    )
+
+  add_test(
+    NAME tests_PythonQtTestMain
+    COMMAND ${Slicer_LAUNCH_COMMAND} $<TARGET_FILE:PythonQtCppTests> tests/PythonQtTestMain
```
usiems: This seems awfully Slicer specific... ;)

CMakeLists.txt
```diff
@@ -0,0 +1,420 @@
...
+
+  if (${PythonQt_Wrap_Qt${qt_wrapped_lib}})
+
+    # Variable expected by PythonQt_QtAll.cpp
+    string(TOUPPER ${qt_wrapped_lib} qt_wrapped_lib_uppercase)
+    ADD_DEFINITIONS(-DPYTHONQT_WITH_${qt_wrapped_lib_uppercase})
+
+    set(file_prefix com_trolltech_qt_${qt_wrapped_lib}/com_trolltech_qt_${qt_wrapped_lib})
+
+    foreach(index RANGE 0 12)
```
usiems: I don't like this loop. We should have the generator generate .cmake files to include, like it generates .pri files.

CMakeLists.txt
usiems, line 1: Do not compile the Qt bindings into PythonQt, make it a separate library, same as with the QMake build system. And use separate CMakeLists.txt files. Probably best to put it into _src_ for PythonQt itself, and into _extensions_ for the bindings similar to qmake (since _generated_cpp_ is only created on-the-fly, but I'm open to other suggestions), and use `add_subdirectory` in the top-level CMakeLists.txt. Perhaps call the bindings project PythonQt_QtAll for now.

In MeVisLab the Qt bindings are dynamically loaded through Python imports, with each module having its own DLL/.so file, so we want to build those separately anyway. It would be easier for us to merge this part back if the bindings are built on their own anyway.

usiems (changes requested):
Also please combine the commits as much as useful. The history of these changes is not that interesting for other users, the only reason to have separate commits is if one might perhaps want to merge or revert only parts of this change.

CMakeLists.txt
```diff
@@ -0,0 +1,275 @@
...
+endif()
+
+#-----------------------------------------------------------------------------
+# Testing
+
+option(BUILD_TESTING "Build the testing tree." OFF)
+include(CTest)
+
+if(BUILD_TESTING)
+  create_test_sourcelist(test_sources PythonQtCppTests.cpp
```
usiems, line 201: I don't understand the use of create_test_sourcelist here. Why is an extra file created here? tests/PythonQtTestMain already contains a main function.

CMakeLists.txt
```diff
@@ -0,0 +1,275 @@
...
+
+#-----------------------------------------------------------------------------
+# Build targets
+add_subdirectory(src)
+if(PythonQt_BUILD_QTALL)
+  add_subdirectory(extensions/PythonQt_QtAll)
+endif()
+
+#-----------------------------------------------------------------------------
+# Testing
```
usiems, line 195: I would like the following part to have its own CMakeLists.txt file in the tests subdirectory.

CMakeLists.txt
```diff
@@ -0,0 +1,275 @@
...
+    set(qtlib_lowercase ${qtlib_to_wraplib_${qtlib}})
+  endif()
+  set(qtlib_to_wraplib_${qtlib} ${qtlib_lowercase})
+endforeach()
+
+#-----------------------------------------------------------------------------
+# Define PythonQt_Wrap_Qt* options
+option(PythonQt_Wrap_QtAll "Make all Qt components available in python" OFF)
+foreach(qtlib ${qtlibs})
+  OPTION(PythonQt_Wrap_Qt${qtlib_to_wraplib_${qtlib}} "Make all of Qt${qtlib} available in python" OFF)
```
usiems, line 112: I would default to ON here, since PythonQt_QtAll is currently only built if PythonQt_Wrap_QtAll (because of the `if(PythonQt_BUILD_QTALL)` before the add_subdirectory). Users could rather switch off the parts they don't want?
Of course this would make the forcing of the individual library bindings below obsolete.

CMakeLists.txt
```diff
@@ -0,0 +1,275 @@
...
+set(qt_required_components Core Widgets)
+foreach(qtlib ${qtlibs})
+  set(qt_wrapped_lib ${qtlib_to_wraplib_${qtlib}})
+  if(${PythonQt_Wrap_Qt${qt_wrapped_lib}})
+    list(APPEND qt_required_components ${qtlib} ${qt_wrapped_lib_depends_${qt_wrapped_lib}})
+  endif()
+endforeach()
+if(BUILD_TESTING)
+  list(APPEND qt_required_components Test)
+endif()
```
usiems, line 144: This rather belongs to the testing part, not to PythonQt.

src/CMakeLists.txt
```diff
@@ -0,0 +1,128 @@
...
+  PUBLIC
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
+    $<INSTALL_INTERFACE:${PythonQt_INSTALL_INCLUDE_DIR}>
+)
+
+target_link_libraries(PythonQt
+  PUBLIC
+    Python3::Python
+    ${QT_LIBRARIES}
```
usiems, line 114: You are linking here against all Qt components needed by the Qt bindings, but this is PythonQt, which only needs Core and Widgets (and CorePrivate). 
The QT_LIBRARIES are needed by PythonQt_QtAll instead. We should try to do this as cleanly as possible.
Also try to move all CMake code from the top-level CMakeLists.txt that is only needed in extensions/PythonQt_QtAll/CMakeLists.txt into that file (e.g., the part that builds QT_LIBRARIES).

ferdymercury (changes requested):
Awesome PR. With this patch I gave a try to restart PythonQt package on Debian: https://salsa.debian.org/python-team/packages/pythonqt

One problem I see with this PR is that PythonQtConfig.cmake is not installed when calling 'make install', meaning one cannot use `find_package(PythonQt)` from outside projects.

Another issue is it would be good to optionally have SOVERSION in the target_properties of the library, for Debian packaging. See for exampel this patch: https://salsa.debian.org/med-team/civetweb/-/blob/master/debian/patches/soversion?ref_type=heads

CMakeLists.txt
```diff
@@ -0,0 +1,275 @@
...
+if(NOT DEFINED PythonQt_INSTALL_RUNTIME_DIR)
+  set(PythonQt_INSTALL_RUNTIME_DIR bin)
+endif()
+
+if(NOT DEFINED PythonQt_INSTALL_LIBRARY_DIR)
+  set(PythonQt_INSTALL_LIBRARY_DIR lib${LIB_SUFFIX})
+endif()
+
+if(NOT DEFINED PythonQt_INSTALL_ARCHIVE_DIR)
+  set(PythonQt_INSTALL_ARCHIVE_DIR lib${LIB_SUFFIX})
```
ferdymercury, line 74: This section here could be modernized. One could do instead

```
include(GNUInstallDirs)
if(NOT DEFINED PythonQt_INSTALL_LIBRARY_DIR)
  set(PythonQt_INSTALL_LIBRARY_DIR ${CMAKE_INSTALL_LIBDIR})
endif()
```
etc. for runtime...

ferdymercury:
PythonQt is now part of Debian: https://packages.debian.org/sid/libpythonqt-dev

The biggest remaining issue is that PythonQtConfig.cmake is not installed, so you can't `find_package` it.

@jamesobutler could you add it? It's only a few lines of code, see for example
https://github.com/yslai/mathtext/pull/3/commits/0783b279855bb8005c68003df975ae05a14484a0

Issue #363, opened 2026-08-21
Connection to a user-defined slot just uses the callable


usiems:
I had defined a slot on one object with the @QtCore.Slot(object) decorator, and a slot with QtCore.Signal(object) on another, connected these and wanted to emit the signal from another thread, so that a queued connection would be used. But the slot was still called from the other thread.

This is what I used for the connection:
`
source.signal.connect(dest.slot)
`
When I replaced this with
`
source.connect("signal(PythonQtSafeObjectPtr)", dest, "slot(PythonQtSafeObjectPtr)")
`
it worked as intended.

What hat happened? I had made the connect in the other thread. PythonQt didn't "realize" that this is a slot on a object, but only "saw" a Python _callable_ and internally created a PythonQtSignalReceiver. Since it was created in the other thread, it lived in this thread, and no queued connection was used.

In the second version, the connect was done to the slot of the dest object, which lived in the main thread, which made Qt use a queued (asynchronous) connection.

I believe on a connect of the first kind PythonQt should first check if this is an existing slot of an object and use that for the connection instead of creating a PythonQtSignalReceiver for the callable, because this is a reasonable expectation of the user. The same logic should apply for connections from signals to signals.

Note: Since this might change the behavior of existing connections, this should at least be a minor version update.

usiems:
This has mostly been fixed with the fix for issue #362. Not in the way that the signal is connected to the defined slot, but the signal is connected to an internal receiver object that should always live in the same thread as the intended receiver object, so the behavior in regard to the connection type should be the same. The differences that arise from not actually using the defined slot but a receiver proxy should rarely matter in practice.

Since I don't expect to invest more work into this issue, I will close it for now.

PR #364, opened 2026-08-21
Fix signal to callable connects

### LINKED ISSUE #362: Can't connect and disconnect Python callables to signals indefinitely
This is because for each source object which has signals connected to callables, a receiver object is generated (class PythonQtSignalReceiver) by PythonQt, and each time a callable is connected to a signal of this object, a virtual slot is generated on this receiver, which gets its own, consecutive slot ID. Slot IDs seem to be internally restricted to 16bit (effectively 15bit) by the connection record used by Qt. So after slot ID 32767 connects stop working.

It seems we need to reuse unused slot IDs after 32767. I would propose to start again with the initial start ID and increase it until a free ID is found.

There is a small risk that the reused slot IDs lead to the wrong callable being called in certain circumstances. For this to happen, the following things need to happen:

- A signal "A" of an object is connected to a callable: An receiver object is created as a child of the sender object.
- The QObject::destroyed signal is connected to a callable: This leads to the receiver being reparented, since otherwise this signal might not be received before the receiver is deleted (special life-time code is employed in this case).
- The sender object is moved to another thread. The sender and its receiver object now belong to different threads.
- The sender object emits signal "A". Because the connection type is Qt::AutoConnection, a QMetaCallEvent is put into the event queue of the receiver's thread.
- Signal "A" is disconnected from the callable.
- An undeterminate number of connects and disconnects to callables happen on the sender.
- A signal from the sender object is connected to a different callable, which coincidentally gets the same slot ID as the previous receiver slot (since it is free now)
- Events are processed on the receiver's thread: The QMetaCallEvent is passed to the receiver object, and the slot ID matches that of the new receiver slot: In this case the wrong callable is called.

An easy remedy to prevent this case would be to use Qt::DirectConnection instead of Qt::AutoConnection when connecting to callables. Callables would always be called on the thread that emits the signal, which probably is what users would expect anway. But we probably need to think somewhat more about thread-safety.

usiems:
Originally this was only meant to fix #362, but then I noticed that a test in our MeVisLab code base failed after the change because of a faulty assumption regarding connections to user-created slots, and so I entered issue #363 and fixed that, too.

I must admit that I am a little bit uneasy about this change, because with these changes the type of connection used (and the target object) depends on factors that are not readily apparent, while before there could be misconceptions about the targeted object in the connect.

I'm open to suggestions.

It is probably advisable to review the commits separately.

### DIFF
--- a/src/PythonQtSignal.cpp
+++ b/src/PythonQtSignal.cpp
@@ -233,18 +233,71 @@ static PyObject* PythonQtSignalFunction_typeName(PythonQtSignalFunctionObject* t
   return PythonQtMemberFunction_typeName(type->m_ml);
 }
 
+// Find out if we can connect directly to the given Python object, without creating a separate receiver.
+// If yes, provide target object and signature for the connect call.
+static bool extractSignalTarget(PyObject* object, QObject*& targetObj, QByteArray& targetSignature)
+{
+  static PyObject* qtSlots = PyUnicode_FromString("_qtSlots");
+  if (PyObject_TypeCheck(object, &PythonQtSignalFunction_Type)) {
+    PythonQtSignalFunctionObject* type = (PythonQtSignalFunctionObject*)object;
+    PythonQtInstanceWrapper* self = (PythonQtInstanceWrapper*)type->m_self;
+    if (self->_obj) {
+      // connecting to another signal
+      targetObj = self->_obj;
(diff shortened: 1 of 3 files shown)


usiems:
I had a discussion with @florianlink about this, and we decided that this needs to be done differently. I will probably only fix #362, #363 is a lot more work to get it right.

PR #365, opened 2026-08-28
Create a separate receiver object for each signal/callable connect

### LINKED ISSUE #362: Can't connect and disconnect Python callables to signals indefinitely

usiems:
This way we can't run out of slot IDs.
This fixes #362

This also enables us to associate the receiver with the instance object of the callable (if it is a method of a QObject-derived class) instead of the sender, so that it is associated with the correct thread, which is important for the AutoConnection used.
In a way this also fixes the problem in #363, because now the receiver object should always live in the same thread as the self object of the callable. The connection to a callable doesn't use the actual slot defined by a slot decorator, but the effect should be largely the same.

Note that this might change the behavior of signal/callable connections when threads are used, so users need to test if their uses still work for them.

### DIFF
--- a/src/PythonQt.cpp
+++ b/src/PythonQt.cpp
@@ -34,7 +34,6 @@
 /*!
 // \file    PythonQt.cpp
 // \author  Florian Link
-// \author  Last changed by $Author: florian $
 // \date    2006-05
 */
 //----------------------------------------------------------------------------------
@@ -324,8 +323,7 @@ void PythonQt::init(int flags, const QByteArray& pythonQtModuleName)
 void PythonQt::cleanup()
 {
   if (_self) {
-    // Remove signal handlers in advance, since destroying them calls back into
-    // PythonQt::priv()->removeSignalEmitter()
+    // Remove all created signal receivers
     _self->removeSignalHandlers();
 
     delete _self;
@@ -879,65 +877,34 @@ PyObject* PythonQtPrivate::createNewPythonQtEnumWrapper(const char* enumName, Py
   return result;
 }
 
-PythonQtSignalReceiver* PythonQt::getSignalReceiver(QObject* obj)
-{
-  PythonQtSignalReceiver* r = _p->_signalReceivers[obj];
-  if (!r) {
-    r = new PythonQtSignalReceiver(obj);
-    _p->_signalReceivers.insert(obj, r);
-  }
-  return r;
-}
-
 bool PythonQt::addSignalHandler(QObject* obj, const char* signal, PyObject* module, const QString& objectname)
 {
-  bool flag = false;
   PythonQtObjectPtr callable = lookupCallable(module, objectname);
   if (callable) {
-    PythonQtSignalReceiver* r = getSignalReceiver(obj);
-    flag = r->addSignalHandler(signal, callable);
-    if (!flag) {
-      // signal not found
-    }
-  } else {
-    // callable not found
+    return _p->addSignalHandler(obj, signal, callable);
   }
-  return flag;
+  // callable not found
+  return false;
 }
 
-bool PythonQt::addSignalHandler(QObject* obj, const char* signal, PyObject* receiver)
+bool PythonQt::addSignalHandler(QObject* obj, const char* signal, PyObject* callable)
 {
-  bool flag = false;
-  PythonQtSignalReceiver* r = getSignalReceiver(obj);
-  if (r) {
-    flag = r->addSignalHandler(signal, receiver);
-  }
-  return flag;
+  return _p->addSignalHandler(obj, signal, callable);
 }
 
 bool PythonQt::removeSignalHandler(QObject* obj, const char* signal, PyObject* module, const QString& objectname)
 {
-  bool flag = false;
   PythonQtObjectPtr callable = lookupCallable(module, objectname);
   if (callable) {
-    PythonQtSignalReceiver* r = _p->_signalReceivers[obj];
-    if (r) {
-      flag = r->removeSignalHandler(signal, callable);
-    }
-  } else {
-    // callable not found
+    return _p->removeSignalHandler(obj, signal, callable);
   }
-  return flag;
+  // callable not found
+  return false;
 }
 
-bool PythonQt::removeSignalHandler(QObject* obj, const char* signal, PyObject* receiver)
+bool PythonQt::removeSignalHandler(QObject* obj, const char* signal, PyObject* callable)
 {
-  bool flag = false;
-  PythonQtSignalReceiver* r = _p->_signalReceivers[obj];
-  if (r) {
-    flag = r->removeSignalHandler(signal, receiver);
-  }
-  return flag;
+  return _p->removeSignalHandler(obj, signal, callable);
 }
 
 PythonQtObjectPtr PythonQt::lookupCallable(PyObject* module, const QString& name)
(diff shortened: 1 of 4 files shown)


src/PythonQt.cpp
```diff
@@ -879,65 +877,36 @@ PyObject* PythonQtPrivate::createNewPythonQtEnumWrapper(const char* enumName, Py
...
-  bool flag = false;
   PythonQtObjectPtr callable = lookupCallable(module, objectname);
   if (callable) {
-    PythonQtSignalReceiver* r = getSignalReceiver(obj);
-    flag = r->addSignalHandler(signal, callable);
-    if (!flag) {
-      // signal not found
-    }
+    return _p->addSignalHandler(obj, signal, callable);
   } else {
```
mrbean-bremen: no sense in the else part, just move the comment down

src/PythonQt.cpp
```diff
@@ -879,65 +877,36 @@ PyObject* PythonQtPrivate::createNewPythonQtEnumWrapper(const char* enumName, Py
...
-    PythonQtSignalReceiver* r = _p->_signalReceivers[obj];
-    if (r) {
-      flag = r->removeSignalHandler(signal, callable);
-    }
+    return _p->removeSignalHandler(obj, signal, callable);
   } else {
     // callable not found
   }
-  return flag;
+  return false;
```
mrbean-bremen, line 902: same here

src/PythonQt.cpp
```diff
@@ -2012,6 +1975,57 @@ PythonQtClassInfo* PythonQtPrivate::lookupClassInfoAndCreateIfNotPresent(const c
...
+bool PythonQtPrivate::addSignalHandler(QObject* sender, const char* signal, PyObject* callable)
+{
+  // Note: It is assumed that the GIL is held when this is called
+  bool ok = false;
+  int sigId = PythonQtSignalReceiver::getSignalIndex(sender, signal);
+  if (sigId >= 0) {
+    // create PythonQtMethodInfo from signal
+    auto* receiver = new PythonQtSignalReceiver(sender, sigId, callable);
+    _signalReceivers.insert(SignalKey(sender, sigId), receiver);
+    ok = true;
```
mrbean-bremen: `return true;`

src/PythonQt.cpp
```diff
@@ -2012,6 +1975,57 @@ PythonQtClassInfo* PythonQtPrivate::lookupClassInfoAndCreateIfNotPresent(const c
...
+  }
+  return ok;
+}
+
+bool PythonQtPrivate::removeSignalHandler(QObject* sender, const char* signal, PyObject* callable)
+{
+  // Note: It is assumed that the GIL is held when this is called
```
mrbean-bremen: Can't you just add `PYTHONQT_GIL_SCOPE` here? That would do nothing is the GIL is already held (same above).

src/PythonQtSignalReceiver.h
```diff
@@ -56,45 +55,30 @@ class PythonQtClassInfo;
...
-
   //! get the signals parameter info
   const PythonQtMethodInfo* methodInfo() const { return _methodInfo; }
 
   //! call the python callable with the given arguments (as defined in methodInfo)
   void call(void** arguments) const;
 
-  //! check if it is the same signal target
-  bool isSame(int signalId, PyObject* callable) const;
+  //! check if it is this targets the same callable
```
mrbean-bremen: `if this targets the same callable`?

src/PythonQtSignalReceiver.h
```diff
@@ -106,42 +90,43 @@ class PythonQtSignalReceiverBase : public QObject
...
-  bool addSignalHandler(const char* signal, PyObject* callable);
+  //! Returns the signal sender for which this receiver was created.
+  QObject* sender() const { return _sender; }
 
-  //! remove a signal handler for given callable (or all callables on that signal if callable is NULL)
-  bool removeSignalHandler(const char* signal, PyObject* callable = nullptr);
+  //! Returns the signal ID for which this receiver was created.
+  int signalId() const { return _signalId; }
+
+  //! Check if this object targets the same callable given as argument.
```
mrbean-bremen: -> targets the callable...

mrbean-bremen (approved):
Looks good as far as I can see (though I may miss some possible implications)

florianlink:
1. removeSignalHandler now removes all matches for a callable
The old code breakd after removing the first matching target; the new loop removes every receiver whose isSameCallable(callable) returns true:

So connect(signal, f); connect(signal, f); disconnect(signal, f) now disconnects both, where previously it disconnected one. This actually matches Qt's native disconnect semantics (removes all matching connections), so it's arguably more correct — but it's a behavior change worth documenting in the PR description / changelog since the author already flagged thread-related behavior changes.

2. Only destroyed(QObject*) is watched now
The old code tracked both destroyed() and destroyed(QObject*) via _destroyedSignal1Id/_destroyedSignal2Id. The new code only connects to destroyed(QObject*). This is fine for modern Qt (the arg variant is the canonical one), but it's a subtle narrowing — if anything on the QObject::staticMetaObject ever emits the no-arg form in some legacy/Qt-version scenario, the receiver wouldn't self-clean. Low risk, just noting.

usiems:
> 1. removeSignalHandler now removes all matches for a callable
>        The old code breakd after removing the first matching target; the new loop removes every receiver whose isSameCallable(callable) returns true:
> 
> 
> So connect(signal, f); connect(signal, f); disconnect(signal, f) now disconnects both, where previously it disconnected one. This actually matches Qt's native disconnect semantics (removes all matching connections), so it's arguably more correct — but it's a behavior change worth documenting in the PR description / changelog since the author already flagged thread-related behavior changes.

Thanks for the hint! I could change it back if you would feel better about it, otherwise I will just document the change.

>  2. Only destroyed(QObject*) is watched now
>        The old code tracked both destroyed() and destroyed(QObject*) via _destroyedSignal1Id/_destroyedSignal2Id. The new code only connects to destroyed(QObject*). This is fine for modern Qt (the arg variant is the canonical one), but it's a subtle narrowing — if anything on the QObject::staticMetaObject ever emits the no-arg form in some legacy/Qt-version scenario, the receiver wouldn't self-clean. Low risk, just noting.

I looked it up: The variant with the QObject* argument exists at least since Qt 5.6 (I would guess since Qt 5.0), and we don't support older Qt versions anymore since PythonQt 4.

